PR #89, opened 2021-12-11
Test framework

drowe67:
Ctests to support TNC modem and sound card integrtaion.

# TODO

1. - [x] GitHub actions to run tests automatically on push
1. - [x] Fix 001 tx and rx in pipe mode
1. - [x] Get 001 running in audio card mode as well
1. - [x] Add meaningful names for used modes in python 
1. - [x] freedv_api.h defines/constants file for Python
1. - [x] 48<->8 resamplers
1. - [ ] ~~Display Bit error rate (ber) for python test_rx.py~~ _DJ2LS: deleted, because no reason for doing this yet_
1. - [ ] ~~Transmit codec2 "testframes" from python test_tx.py~~ _DJ2LS: deleted, because no reason for doing this yet_
1. - [x] Create a buffer management for NIN and reading fixed audio chunks  instead of reading NIN-chunks directly from pyaudio loop
1. - [ ] Check resampler and buffer code to pythonize it a little bit more.
1. - [ ] Add tests for pyAudio callback mode

# Instructions

1. Install:
   ```
   cd FreeDATA
   mkdir build
   cd build
   cmake -DCODEC2_BUILD_DIR=$HOME/codec2/build_linux ..
   ```
2. List available tests:
   ```
   ctest -N
   Test project /home/david/FreeDATA/build
   Test #1: 000_audio_tests
   Test #2: 001_highsnr_stdio_audio

   Total Tests: 2
   ```
3. Run tests:
   ```
   ctest --output-on-failure
   ```
4. Run tests verbosely:
   ```
   ctest -V
   ```


DJ2LS:
@drowe67 I edited your todo list and added check buttons and point 6. and 7.

test/001_highsnr_stdio_audio/test_rx.py
```diff
@@ -0,0 +1,149 @@
...
+
+c_lib.freedv_get_n_max_modem_samples.argtype = [c_void_p]
+c_lib.freedv_get_n_max_modem_samples.restype = c_int
+
+c_lib.freedv_set_frames_per_burst.argtype = [c_void_p, c_int]
+c_lib.freedv_set_frames_per_burst.restype = c_void_p
+      
+c_lib.freedv_get_rx_status.argtype = [c_void_p]
+c_lib.freedv_get_rx_status.restype = c_int      
+
```
drowe67, line 80: @DJ2LS these are likely to be common to many programs. Could we put these "function headers" (using a C term) into a common file/module that we can import into many programs?

test/002_highsnr_ping_pong/ping.py
```diff
@@ -0,0 +1,199 @@
...
+        time.sleep(0.01)
+   
+        nin = c_lib.freedv_nin(freedv)
+        nin_converted = int(nin*(AUDIO_SAMPLE_RATE_RX/MODEM_SAMPLE_RATE))
+        if DEBUGGING_MODE == True:
+            print("-----------------------------")
+            print("NIN:  " + str(nin) + " [ " + str(nin_converted) + " ]")
+        
+        data_in = stream_rx.read(nin_converted,  exception_on_overflow = False)  
+        data_in = data_in.rstrip(b'\x00')
```
drowe67, line 103: I'd suggest a design that [reads a constant buffer size](https://github.com/drowe67/codec2/pull/265/files#diff-194711ca5dcb53842b73b19af641619f1065e2390e8b302218adaa9e85aff4e5) from the audio card, then uses that to fill a buffer up to `nin` samples.  That's just a gut feel, based on experience I guess (and mistrust of Linux audio!).

test/001_highsnr_stdio_audio/test_rx.py
DJ2LS: seems to be solved with latest commit

DJ2LS: functions can be accessed by `codec2.api.freedv_open(12)` for example.

test/002_highsnr_ping_pong/ping.py
DJ2LS: I think I already did this with my multimode test :thinking: . I need to change it so it fits to our tests. I will do this the next days.

drowe67:
@DJ2LS is test 000 a good match/relevant for the resampling you are doing in FreeDATA?  If so I can work out a way to check the results.  There seem to be a few different resamplers in [sinustest.py](https://github.com/DJ2LS/FreeDATA/pull/89/files#diff-b3ca39dd0daf0acd655afef947307e8eefa4c76e06d247236c2cbfc55d859df4)

DJ2LS:
No, it is not relevant I think. It was for testing different resamplers by checking the resampled output "by ear".
We can remove it in my opinion as there is no further benefit for the other tests

drowe67:
@DJ2LS I changed the test naming system slightly to:
```
ctest -V -R 001_highsnr_stdio_C_P
```
As using <> meant we needed;
```
ctest -V -R "001_highsnr_stdio_C<>P"
```
which confused me while running individual tests :slightly_smiling_face:

test/002_highsnr_ping_pong/ping.py
DJ2LS: I will also try doing this within our single mode tests. We are also working out a good example/demo for learning codec2 by doing this.
I'll add this to the todo list.

test/001_highsnr_stdio_audio/test_tx.py
```diff
@@ -0,0 +1,141 @@
...
+DELAY_BETWEEN_BURSTS = args.DELAY_BETWEEN_BURSTS/1000
+AUDIO_OUTPUT_DEVICE = args.AUDIO_OUTPUT
+
+if args.FREEDV_MODE == "datac0":
+    MODE = codec2.api.FREEDV_MODE_DATAC0
+if args.FREEDV_MODE == "datac1":
+    MODE = codec2.api.FREEDV_MODE_DATAC1
+if args.FREEDV_MODE == "datac1":
+    MODE = codec2.api.FREEDV_MODE_DATAC3
+
```
drowe67, line 39: @DJ2LS :point_up: - is it OK if we move to meaningful string based mode names?  There is probably a more pythonic way to convert the strings to constants.  :thinking: - this is also code that would be useful to place in a shared library.

DJ2LS: Yes I agree!
Maybe we could do this with [enum](https://docs.python.org/3/howto/enum.html). I'll have a look the next days and add it to the todo list.
Are you talking about the codec2 api and something like this?
` mode = freedv_get_mode('datac0')`

DJ2LS: This should be solvable by moving it to the codec2 module where we are storing the header information 🤔

drowe67:
Unfortunately I couldn't install alsa modules in the ctest github workflow:
```
sudo modprobe snd-aloop index=1,2 enable=1,1 pcm_substreams=1,1 id=CHAT1,CHAT2
```
It might be we can't install kernel modules, or maybe I just need to find the right package with `snd-aloop`

test/001_highsnr_stdio_audio/test_tx.py
drowe67: Yes a function like that would be very useful.

drowe67:
> No, it is not relevant I think. It was for testing different resamplers by checking the resampled output "by ear". We can remove it in my opinion as there is no further benefit for the other tests.

Cool, I'll remove it.

> But I'm not sure if the resamplers are from equal quality. I could test this maybe the next days with testframes and bit error rate?

They should give very similar behavior, just statistical variations in BER.  A test around the chosen resampler might be useful, just to make sure we are using it correctly and that there are no issues with our impleentation.  This could be a sine wave test, or a modem/BER test.

DJ2LS:
@drowe67 I changed the name of the tests a little bit so we can differ between singlemode and multimode tests:
**multimode test run**
`ctest -V -R 001_highsnr_stdio_C_P_MM`

**singlemode test run**
`ctest -V -R 001_highsnr_stdio_C_P_SM`

drowe67:
@DJ2LS the new multimode test looks good to me :+1: 

It is quite slow to run, I also noticed this in the [C version](https://github.com/drowe67/codec2/pull/265#issuecomment-991392854).  The acquisition code is very slow, although with two parallel receivers it was about 18% of real time on my machine (two receivers) which should be OK.  If this becomes a problem I can do some work on optimisation of the C code.  Excessive CPU load could also cause a problem with real time operation (and maybe sound buffer I/O), so we should monitor this somehow.

In the TNC, do we need to run 3 parallel receivers?  I had assumed the default use case was the current payload mode (e.g. datac1 or datac3), plus datac0 to handle unexpected signalling frames - two receivers in total.

test/001_highsnr_stdio_audio/test_virtual.sh
```diff
@@ -0,0 +1,25 @@
...
+check_alsa_loopback
+
+RX_LOG=$(mktemp)
+
+# make sure all child processes are killed when we exit
+trap 'jobs -p | xargs -r kill' EXIT
+
+python3 test_rx.py --mode datac0 --frames 2 --bursts 1 --audiodev 4 --debug 2>&1 | tee ${RX_LOG} &
+sleep 1
+python3 test_tx.py --mode datac0 --frames 2 --bursts 1 --audiodev 5
```
drowe67, line 23: rather than a sleep, might be better to wait on a string from `test_rx.py`

drowe67:
Perhaps now we have `test_rx.py` and `test_tx.py` running with the audio cards, we might be able to run them over radios :thinking: That would be cool :sunglasses:

test/001_highsnr_stdio_audio/test_virtual.sh
drowe67: One problem here is the audiodev numbers won't be consistent across machines.  I couldn't find the loopback device logic names in the PortAudio device list.

test/001_highsnr_stdio_audio/test_multimode_rx.py
```diff
@@ -0,0 +1,161 @@
...
+receive = True
+
+BUF_SIZE = 160
+
+while receive and time.time() < timeout:
+    if AUDIO_INPUT_DEVICE != -1: 
+        buf_size_converted = int(BUF_SIZE*(AUDIO_SAMPLE_RATE_RX/MODEM_SAMPLE_RATE))
+        data_in = stream_rx.read(buf_size_converted,  exception_on_overflow=False)
+        data_in = audioop.ratecv(data_in,2,1,AUDIO_SAMPLE_RATE_RX, MODEM_SAMPLE_RATE, None)
+        data_in = data_in[0].rstrip(b'\x00')
```
drowe67, line 102: @DJ2LS - I have a strange bug where I'm getting `len(data_in) = 2` at this point, it's 1920 as expected after L100.

drowe67: BTW, why do we need `rstrip`?

test/001_highsnr_stdio_audio/test_virtual.sh
drowe67: Actually I think I'm seeing some of problems you have experienced with `test_rx.py` in sound card mode.  Sometimes it doesn't receive the frames, and returns the error flag in the status returned from the modem. I'll dig into this a little more (my) tomorrow.

test/codec2.py
```diff
@@ -0,0 +1,105 @@
...
+    "-BST",
+    "E---",
+    "E--T",
+    "E-S-",
+    "E-ST",
+    "EB--",
+    "EB-T",
+    "EBS-",
+    "EBST"]
+
```
drowe67, line 100: @DJ2LS pls feel to refactor this if it's not Pythonic :slightly_smiling_face:

drowe67:
OK I can reproduce the audio I/O issues:

1. Using pipes:
   ```
   python3 test_tx.py --mode datac0 --frames 2 --bursts 5  --delay 500 | python3 test_rx.py --mode datac0 --frames 2 --bursts 5
   <snip>
   RECEIVED BURSTS: 5 RECEIVED FRAMES: 10 RX_ERRORS: 0
   ```
2. Virtual sound card:
   ```
   ./test_virtual.sh
   <snip>
   RECEIVED BURSTS: 2 RECEIVED FRAMES: 4 RX_ERRORS: 2
   ```
   
   ```

drowe67:
@DJ2LS I'm pretty sure it's the resampler.  The code for the audioop resampling is [here](https://github.com/python/cpython/blob/bb3e0c240bc60fe08d332ff5955d54197f79751c/Modules/audioop.c#L1477).  It's not a proper FIR filter resampler, and there are some issues about preserving the resampler state between calls.

Changing AUDIO_SAMPLE_RATE to 8000 Hz seems to fix the errors.  When I play to real sound card it sounds OK on my laptop.  We may have problems with some sounds cards however, I'm not sure how PortAudio handles sample rates internally.

I use libresample for resampling in freedv-gui, here is a [C example](https://github.com/drowe67/codec2/blob/master/misc/tsrc.c)

drowe67:
@DJ2LS in https://github.com/DJ2LS/FreeDATA/pull/89/commits/3e1dc4ee1cdfb309bec1125aaa314fc4cae495d1 I redesigned the buffers management in `test_rx.py` again (sorry!).  The previous design had a constantly growing buffer that I suspect meant lots of memory system activity.  I think this might have been the issue with the very slow running speed.

The current implementation uses the design in [freedv_data_c0c1_rx.c](https://github.com/drowe67/codec2/blob/dr-tnc/demo/freedv_datac0c1_rx.c). I suspect C is a better language for this low level work, however my Python knowledge isn't great, so their might be a better way ...

drowe67:
Spoke too soon about the multimode test - it's not detecting multiple frames :thinking: It is running nice and fast though.

drowe67:
> Spoke too soon about the multimode test - it's not detecting multiple frames thinking It is running nice and fast though.

Actually I think it was running, it's just the logging is tricky with multiple parallel modems, as they operate at different rates.  We really need three different consoles.  Also I'm not sure we want to print too much to the console when trying to run in real time.

drowe67:
@DJ2LS I remembered I had a 48<->8kHz resampler in codec2 (well, some years back in the git history). Now I have [it running](https://github.com/drowe67/codec2/blob/dr-tnc/src/fdmdv.c#L1911) in libcodec2 dr-tnc branch.  Next step is to write a Python unit test, similar to the [C unit test](https://github.com/drowe67/codec2/blob/dr-tnc/unittest/t48_8_short.c).

drowe67:
> Ah great blush can the resampler used like the other functions of the codec2 api? 

Yes, I'm working up an interface for it:
```
api.fdmdv_8_to_48_short.argtype = [c_void_p, c_void_p, c_int]
api.fdmdv_48_to_8_short.argtype = [c_void_p, c_void_p, c_int]
```
It has a few restrictions, but it works in "streaming" mode which is just what we want.  Most other Python resamplers seem to work on file sized arrays.

drowe67:
@DJ2LS I have `test_tx.py` and `test_rx.py` running at 48 kHz now, even when running in pipe mode.  I made this choice to give the resampler a good work out without the complication of also having to run Python audio.  I've modified the C<->P ctests, using `sox` to interface between the 8 and 48 kHz sample rates.

drowe67:
@DJ2LS I made some progress today on the resampler, and the tests work OK in pipe mode.  However I have some odd behaivour with virtual and real sound cards, for example:
```
python3 test_tx.py --mode datac0 --bursts 2 --framesperburst 1 --delay 500 --audiodev 12
```
In this test (which plays out of my pulseaudio sound device), the interburst silence is too short :thinking:

drowe67:
100 packets OTC using Python Audio I/O :champagne: 

Laptop1->Rigblaster->FT817->attenuators->IC7200 USB->Laptop2 

Laptop 2 Rx:
```
python3 test_rx.py --mode datac0 --bursts 100 --framesperburst 1 --audiodev 8 --timeout 600 --debug
<snip>
RECEIVED BURSTS: 100 RECEIVED FRAMES: 100 RX_ERRORS: 0
```

Then on Laptop 1 Tx:
```
python3 test_tx.py --mode datac0 --bursts 100 --framesperburst 1 --audiodev 2 --delay 250
```

drowe67:
Here's a 7 minute (!) OTC test using 5 frame bursts data datac1:
```
RECEIVED BURSTS: 20 RECEIVED FRAMES: 100 RX_ERRORS: 0
```
I make that 51,000 bytes.  Running for 7 minutes error free sure is a confidence booster :slightly_smiling_face:

test/001_highsnr_stdio_audio/test_rx.py
```diff
@@ -0,0 +1,180 @@
...
+   
+N_BURSTS = args.N_BURSTS
+N_FRAMES_PER_BURST = args.N_FRAMES_PER_BURST
+AUDIO_INPUT_DEVICE = args.AUDIO_INPUT_DEVICE
+MODE = codec2.FREEDV_MODE[args.FREEDV_MODE].value
+DEBUGGING_MODE = args.DEBUGGING_MODE
+TIMEOUT = args.TIMEOUT
+
+# AUDIO PARAMETERS
+AUDIO_FRAMES_PER_BUFFER = 2400*2  # <- consider increasing if you get nread_exceptions > 0
```
drowe67, line 50: I needed to fiddle with this to stop audio break up :point_up: Note it also needs to be a multiple of 48/8=6 for the resampler to work.

test/001_highsnr_stdio_audio/test_rx.py
```diff
@@ -0,0 +1,180 @@
...
+nread_exceptions = 0
+timeout = time.time() + TIMEOUT
+receive = True
+audio_buffer = codec2.audio_buffer(AUDIO_FRAMES_PER_BUFFER*2)
+resampler = codec2.resampler()
+
+# Copy received 48 kHz to a file.  Listen to this file with:
+#   aplay -r 48000 -f S16_LE rx48.raw
+# Corruption of this file is a good way to detect audio card issues
+frx = open("rx48.raw", mode='wb')
```
drowe67, line 113: :point_up: monitoring what the Rx receives is how I spotted the audio break up.  I modified one of the tests to play a speech wave file over the virtual card instead of modem tones - that made it really obvious.

test/001_highsnr_stdio_audio/test_rx.py
```diff
@@ -0,0 +1,180 @@
...
+#   aplay -r 48000 -f S16_LE rx48.raw
+# Corruption of this file is a good way to detect audio card issues
+frx = open("rx48.raw", mode='wb')    
+
+# initial number of samples we need
+nin = codec2.api.freedv_nin(freedv)
+while receive and time.time() < timeout:
+    if AUDIO_INPUT_DEVICE != -1:
+        try:
+            data_in48k = stream_rx.read(AUDIO_FRAMES_PER_BUFFER, exception_on_overflow = True)
```
drowe67, line 120: `read()` was actually throwing exceptions, we just had them disabled.   As this is a tricky problem I've put logic in to signal any issues should (lol - when) they happen again.....

test/001_highsnr_stdio_audio/test_tx.py
```diff
@@ -0,0 +1,163 @@
...
+    #print(f"samples_delay: {samples_delay} DELAY_BETWEEN_BURSTS: {DELAY_BETWEEN_BURSTS}", file=sys.stderr)
+
+    # resample up to 48k (resampler works on np.int16)
+    x = np.frombuffer(txbuffer, dtype=np.int16)
+    txbuffer_48k = resampler.resample8_to_48(x)
+    
+    # check if we want to use an audio device or stdout
+    if AUDIO_OUTPUT_DEVICE != -1:
+        # Gotcha: we have to convert from np.int16 to Python "bytes"
+        stream_tx.write(txbuffer_48k.tobytes())
```
drowe67, line 151: This caught me for a while - `write()` doesn't actually choke when you pass it a `np.int16` array, but it's corrupted on playback.  I tracked this one down by playing a test sinewave using `test_pa.py`

All this moving back and forth between `bytes` and `np.int16` is a candidate for refactoring.  I'm not sure if it's too messy.  It's just I understand numpy a bit better than these `byte` thingies.

drowe67:
Multimode tests:
Rx connected to IC7200:
```
python3 test_multimode_rx.py --audiodev 8 --debug --timeout 600 --bursts 40
<snip>
DATAC0: 40/40 DATAC1: 40/40 DATAC3: 40/40
```
Tx connected to FT817:
```
python3 test_multimode_tx.py --audiodev 2 --burst 40
```

DJ2LS:
Updated the todo list

drowe67:
@DJ2LS the Python [array type](https://docs.python.org/3/library/array.html) looks like a good choice for the speech buffer work, it even has built in C pointer support.  I found the use of `bytes` confusing as it needed lots of *2 factors - this is something the programming language should take care of!

DJ2LS:
> Interesting thread on PyAudio input overflow issues: https://stackoverflow.com/questions/10733903/pyaudio-input-overflowed.
> 
> 
> 
> I think PyAudio closes the stream when the input buffer overflows. This is undesirable - we would like to detect and count overflows (to detect serious problems), but keep the stream going. If we miss the occasional buffer the protocol will recover any lost packets.  On a general purpose operating system, it is reasonable to expect the occasional loss of buffer (click) in the audio.  
> 
> 
> 
> In a real world TNC application - the Rx stream could be running indefinitely, listening for incoming stations for weeks without a restart. So I guess a requirement is that it  doesn't stop the input stream under any circumstances.
> 
> 
> 
> The virtual sound cards tests that use aplay/arecord works just fine, but I get occasional fails on the PyAudio versions when the input overflow exception fires.  I suspect this will also pop up on real audio cards. So in future we may need a little more refactoring to handle input overflows better. 
> 
> 
> 
> In [freedv-gui](https://github.com/drowe67/freedv-gui/blob/master/src/main.cpp#L2870) we use the callback method, which is also available in PyAudio. a thread safe FIFO is then used to send samples to the thread that processes the data.

The good thing is, I already did some tests with pyaudio callback mode. From a theoretical point of view it shouldn't be a big problem using the callback mode. Specially with your new buffer management it should be working much better than before. 
Using the callback mode could be a candidate for another test.

DJ2LS:
> @DJ2LS the Python [array type](https://docs.python.org/3/library/array.html) looks like a good choice for the speech buffer work, it even has built in C pointer support.  I found the use of `bytes` confusing as it needed lots of *2 factors - this is something the programming language should take care of!

Ah Great, thanks for the notice! I have been already working with this in parts. I think this library is worth looking a little bit closer at as part of pythonising the current code a little bit 🙂

When we are having a look at the wiki of python we should select version 3.6 and not the latest to keep a compatibility to Ubuntu 20.04 LTS which ships python 3.6

DJ2LS:
@drowe67 I moved codec2.py to "FreeDATA/tnc/codec2.py" . This is a preparation for the general usage of the codec2 module. I want to avoid, that we have equal files within test and tnc ( codec2.py ) So we have to change files only once. 

I also added a test with using pyAudio callbacks, but this forces the resambler to crash for some reason. The result is a failing test.

test/001_highsnr_stdio_audio/test_callback_rx.py
```diff
@@ -0,0 +1,178 @@
...
+rx_total_frames = 0
+rx_frames = 0
+rx_bursts = 0
+rx_errors = 0
+nread_exceptions = 0
+timeout = time.time() + TIMEOUT
+receive = True
+audio_buffer = codec2.audio_buffer(AUDIO_FRAMES_PER_BUFFER*2)
+resampler = codec2.resampler()
+
```
drowe67, line 125: @DJ2LS I think this needs to be created before the stream is started, to set up the state of the resampler.

test/001_highsnr_stdio_audio/test_callback_rx.py
```diff
@@ -0,0 +1,178 @@
...
+# make sure our resampler will work
+assert (AUDIO_SAMPLE_RATE_RX / MODEM_SAMPLE_RATE) == codec2.api.FDMDV_OS_48
+
+
+# ------------------------------------------------ PYAUDIO CALLBACK
+def callback(data_in48k, frame_count, time_info, status):
+    x = np.frombuffer(data_in48k, dtype=np.int16)
+    x.tofile(frx)    
+    x = resampler.resample48_to_8(x)
+    audio_buffer.push(x)
```
drowe67, line 63: I don't think audio buffer is thread safe,  e.g. the callback and main thread could be making audio_buffer calls at the same time.  So we might need some syncronisation objects like mutexes (of the Python equivalent) around the audio buffer push and pop functions.

The [codec2_fifo](https://github.com/drowe67/codec2/blob/master/src/codec2_fifo.h) design is thread safe by design, but we'd need to re-write it in Python, or call the C version from Python.

test/001_highsnr_stdio_audio/test_callback_rx.py
DJ2LS: I changed the order of this part also this afternoon and now again, but the resampler still crashes. I would'nt be surprised if the problem is pyAudio's callback data which differs from the normal blocking mode. I had some curious behavior when using the callback function some weeks ago for writing data to the audio stream :thinking: But I'm not sure if we really need the callback mode. The next days will maybe clarify this, I think.

test/001_highsnr_stdio_audio/test_callback_rx.py
DJ2LS: I'm surprised again about the hidden features of codec2 :) However, as I wrote above, I'm not sure if we really need the callback mode. Maybe its enough running the entire audio part within a own thread. 

This is also an area I need to learn more to get a better understanding.

drowe67:
Ok I found the problem with using the resampler from the callback, the memory addresses are being corrupted across the Python<-> boundary;
```
calling .... out8: 0x7fcbb8001520 pin48: 0x7fcbb80091c0 <-- resample48_to_8()
hello! out8k 0x7fcbb8001520 in48k 0xffffffffb80091c0 n 800 <---  C fdmdv_48_to_8_short()
Segmentation fault (core dumped)
```
I think this is due to me still on the Python<->C learning curve.... I suspect the proper solution is to use the Python ctypes library in the resampler.  I will look into this ...

test/001_highsnr_stdio_audio/test_callback_rx.py
drowe67: From reading a few stack exchange threads, looking at Horus GUI, and my experience with PortAudio on freedv-gui .. I get the feeling that the callback model might be best.

test/001_highsnr_stdio_audio/test_callback_rx.py
drowe67: Yes I have been going down the same path ... :thinking:

drowe67:
Here's the resampler fix and a first pass at `test_callback_rx.py` working: https://github.com/DJ2LS/FreeDATA/pull/92

PR #92, opened 2021-12-19
PyAudio rx callback and Python<->C Resampler

drowe67:
Prototyping the callback model, and fixes to C<->Python resampler interface.  

@DJ2LS - I've put this in a separate PR to highlight the changes.

The ctests made it really easy to test the resampler changes :slightly_smiling_face:


test/001_highsnr_stdio_audio/test_callback_rx.py
```diff
@@ -0,0 +1,186 @@
...
+# Corruption of this file is a good way to detect audio card issues
+frx = open("rx48_callback.raw", mode='wb')    
+
+# ------------------------------------------------ PYAUDIO CALLBACK
+def callback(data_in48k, frame_count, time_info, status):
+    global total_n_bytes
+    global rx_total_frames
+    global rx_frames
+    global rx_bursts
+    global receive
```
drowe67, line 95: :point_up: all these globals are messy, maybe put this all in class like `horus-gui`

test/001_highsnr_stdio_audio/test_callback_rx.py
```diff
@@ -0,0 +1,186 @@
...
+
+            if rx_frames == N_FRAMES_PER_BURST:
+                rx_frames = 0
+                rx_bursts = rx_bursts + 1
+                          
+            if rx_bursts == N_BURSTS:
+                receive = False
+    
+    return (None, pyaudio.paContinue)
+
```
drowe67, line 135: :point_up: I'm warming to the idea of putting all the signal processing in the callback thread.

tnc/codec2.py
```diff
@@ -0,0 +1,205 @@
...
+        # length of input vector must be an integer multiple of api.FDMDV_OS_48
+        assert(len(in48) % api.FDMDV_OS_48 == 0)
+
+        # concat filter memory and input samples
+        in48_mem = np.zeros(self.MEM48+len(in48), dtype=np.int16)
+        in48_mem[:self.MEM48] = self.filter_mem48
+        in48_mem[self.MEM48:] = in48
+
+        # In C: pin48=&in48_mem[MEM48]
+        pin48 = byref(np.ctypeslib.as_ctypes(in48_mem), 2*self.MEM48)
```
drowe67, line 179: This line cost me three hours of my life and I want them back :grimacing: !

Ideas for refactoring this class: I think embracing the ctypes library and maybe using `create_string_buffer()` for `in48_mem` here may be the way to go.  Ctypes even has `memcpy` which makes me happy.

I'm finding Python is so "big" compared to C.  So many libraries and data types, and different ways of doing things.

test/001_highsnr_stdio_audio/test_callback_rx.py
DJ2LS: I also did this some weeks ago with some small success. One of the problems was the resampling part and processing time. So we could test again and see how it performs with the new resampler and buffer design.

tnc/codec2.py
DJ2LS: I'm feeling with you! 

I think if we want to do this, we have to think about the dtype=int16 part of numpy. If we are moving away from this and handling the audio data as a byte string, we could maybe make the code a little bit simpler. One part of confusion is the usage of numpy arrays + directly access data via ctypes.

But I would suggest, that this is a last step or maybe a part for another PR as soon as we found a working design for handling the audio part.

And yes, doing things in many different way may be easy in the beginning when starting learning python, because everything is working somehow. But  some when a point is reached, where it causes more confusion and ugly code. For this reason I was thinking about moving to another programming language. But I'm not sure if this is a good idea....

test/001_highsnr_stdio_audio/test_callback_rx.py
DJ2LS: Class :ballot_box_with_check:

test/001_highsnr_stdio_audio/test_callback_rx_outside.py
```diff
@@ -0,0 +1,181 @@
...
+        self.frx = open("rx48_callback.raw", mode='wb')    
+           
+    def callback(self, data_in48k, frame_count, time_info, status):
+        
+        x = np.frombuffer(data_in48k, dtype=np.int16)
+        x.tofile(self.frx)
+        x = self.resampler.resample48_to_8(x)    
+        self.audio_buffer.push(x)
+
+        return (None, pyaudio.paContinue)
```
DJ2LS, line 122: @drowe67 I created this test to determine if we have any differences with decoding inside and outside of the callback. This could become interesting for multimode decoding. 
It could also be possible, we have less memory address problems because of the class :thinking:

drowe67: I'm sure it's possible either way, however with the decoding outside the callback we will need way to prevent two threads accessing the audio buffer at the same time (this will eventually happen with a long enough run).

tnc/codec2.py
drowe67: > But I would suggest, that this is a last step or maybe a part for another PR as soon as we found a working design for handling the audio part.

Yes I agree :+1: No rush on the refactoring.

DJ2LS:
First test runs via virtual sound devices are showing, that placing the decoding part inside the audio callback is performing slightly better than placing the decoding part outside. 
This needs to be verified by testing more and OTC.

test/test_tx.py
```diff
@@ -0,0 +1,171 @@
...
+
+# data binary string
+if args.TESTFRAMES:
+    data_out = bytearray(14)
+    data_out[:1]   = bytes([255])
+    data_out[1:2]  = bytes([1])
+    data_out[2:]  = b'HELLO WORLD'
+    
+else:
+    data_out = b'HELLO WORLD!'
```
DJ2LS, line 87: Option for using the test programs for checking the TNC itself. Just append `--testframes` as additional parameter. This could be used maybe as standard instead of sending a simple b'HELLO WORLD!'

drowe67:
> First test runs via virtual sound devices are showing, that placing the decoding part inside the audio callback is performing slightly better than placing the decoding part outside. This needs to be verified by testing more and OTC.

I thought of one edge case where doing all the processing in the callback might be a problem - if the processing takes longer than the period of one audio buffer.  However lets wait until we see that happen :thinking:

srsampson:
I was wondering about the resampler in codec2. The ```short``` version casts the PCM to float, but I'm wondering if casting it to float, and then dividing bt 32768.0 in order to work with smaller numbers (say -1.0 to 1.0).

I may not be seeing that this is done elsewhere, so just a thought, Steve.

DJ2LS:
This mornings tests are showing, that `highsnr_virtual5_P_P_multi_callback` is performing slightly better with threaded logging. 
However, `highsnr_virtual5_P_P_multi_callback_outside` performs better this morning than `highsnr_virtual5_P_P_multi_callback` .

drowe67:
Sorry I can't see this comment on the PR anymore?

I think we are happy with the normal +/2^15 range of samples, as that's
what the sound cards and modem work on.

On Mon, 2021-12-20 at 15:29 -0800, srsampson wrote

drowe67:
> This mornings tests are showing, that `highsnr_virtual5_P_P_multi_callback` is performing slightly better with threaded logging. However, `highsnr_virtual5_P_P_multi_callback_outside` performs better this morning than `highsnr_virtual5_P_P_multi_callback` .

How did you measure the performance to compare? Ideally they should both pass 100% of the time.

DJ2LS:
> > This mornings tests are showing, that `highsnr_virtual5_P_P_multi_callback` is performing slightly better with threaded logging. However, `highsnr_virtual5_P_P_multi_callback_outside` performs better this morning than `highsnr_virtual5_P_P_multi_callback` .
> 
> How did you measure the performance to compare? Ideally they should both pass 100% of the time.

I run them about 10 times. But better comparing them OTC to avoid problems caused by the virtual audio devices.

DJ2LS:
This mornings OTC test results:

`test_tx` --> `test_rx` /  framesperburst 5, bursts 5, delay 200 / ✅ 
`test_tx` -->` test_callback_rx` /  framesperburst 5, bursts 5, delay 200 / ✅ 
`test_tx` -->` test_callback_rx_outside` /  framesperburst 5, bursts 5, delay 200 / ✅ 
`test_multimode_tx` --> `test_multimode_rx` / framesperburst 5, bursts 5, delay 200 / ✅ 
`test_multimode_tx` --> `test_callback_multmode_rx` / framesperburst 5, bursts 5, delay 200 / ✅ 
`test_multimode_tx` --> `test_callback_multmode_rx_outside` / framesperburst 5, bursts 5, delay 200 / ✅

DJ2LS:
I did a test with placing TX and RX within the same callback. This simplifies the overall audio code, but creates a blocking state for TX audio, if the decoding part for RX is within the callback.
So it seems, we need to place the decoding part outside the callback, for a TNC use case 🤔

DJ2LS:
> I did a test with placing TX and RX within the same callback. This simplifies the overall audio code, but creates a blocking state for TX audio, if the decoding part for RX is within the callback.
> 
> So it seems, we need to place the decoding part outside the callback, for a TNC use case 🤔 

This could be another test.
A test program which is able doing tx and rx within one callback.

drowe67:
> I did a test with placing TX and RX within the same callback. This simplifies the overall audio code, but creates a blocking state for TX audio, if the decoding part for RX is within the callback. So it seems, we need to place the decoding part outside the callback, for a TNC use case thinking

Do we need full duplex Tx and Rx?

PR #157, opened 2022-04-10
Better handle check_callsign changes.

kronenpj:
This updates the way `check_callsign` results are returned. I found that when `check_callsign` failed, it wasn't returning an array, but simply `False`.  Since it wasn't being checked before trying to access element `[1]` of the array, it threw an exception. This makes the return values more consistent and handles the `False` case more gracefully.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -206,7 +206,9 @@ class DATA():
 
 
         #if bytes(bytes_out[1:3]) == self.mycallsign_CRC or bytes(bytes_out[2:4]) == self.mycallsign_CRC or frametype == 200 or frametype == 250:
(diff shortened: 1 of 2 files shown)


kronenpj:
Also inadvertently added the "process all PING packet" change. Sorry.

DJ2LS:
No problem, I'm happy about this pull request and I also learned from it. No one checked my code ( and my failures and design issues ), yet. So your help and feedback is highly welcomed 🙂

DJ2LS:
Oh, the ctests are actually broken. So except an error when merging

kronenpj:
Be careful what you ask for. I've been trying to get better with Python through the pandemic. I may be a nightmare for your code :)
I will make one recommendation from the start. Take a look at `black` and run it against all your Python code. I've found it greatly helps readability.

DJ2LS:
Ah nice, I also had a look at other code validation tools, but never heard about black. I will check this the next days. Thanks

Issue #156, opened 2022-04-10
Have TNC respond to ping without DX Call - Can anybody hear me?


kronenpj:
**Is your feature request related to a problem? Please describe.**
A nice feature of JS8Call allows the operator to send a 'heartbeat' message and any station on frequency has the option to respond with a signal report for that station. It allows an operator to have an idea of how far and how well the station is heard.

**Describe the solution you'd like**
I'd like to propose that
1. A configuration option is created that allows a response to a BEACON or PING.
1. If configured, the TNC responds to a BEACON packet or PING packet without a DX Call.

**Describe alternatives you've considered**
Adding a new packet type isn't necessary as a BEACON currently exists and the current behavior for an un-directed PING is to ignore the packet. However, BEACON isn't responded to and so requires a new response packet. PING has a response packet but is directed. ~~Extending PING appears to be the least disruptive as stations without this change ignore the packet and stations with this change behave as though the PING was directed at them.~~ The responding station's callsign isn't included in the packet. This causes a bit of a conundrum as extending the PING response may cause issues with unmodified stations.

**Additional context**
I would be happy to make an attempt to provide a pull request that extends the TNC code to this end, if it's something you believe would be useful. I wouldn't try to extend the content of the response to include the Signal-to-Noise ratio until a later request.

DJ2LS:
Thanks for the idea! This shouldn't be that hard to implement, we just need to open the "callsign crc filter".

How should we handle the situation, if several stations are receiving the signal and going to respond?  Are there some best practices yet for this? Maybe a random timer?

Thanks, Simon

kronenpj:
A random timer is probably the only way to do it. There will be inevitable collisions but I don't know of another way, unless we do something based on a has of the station's callsign or location. Right now I'm using `time.sleep(randrange(1, 100) / 100.0)` for a 0.01 to 1 second delay.

Until I figure out the configuration option, I added a condition for `frametype == 210` at line 209:
`if helpers.check_callsign(self.mycallsign, bytes(bytes_out[1:3])) or helpers.check_callsign(self.mycallsign, bytes(bytes_out[2:4])) or frametype == 200 or frametype == 210 or frametype == 250:`
This should address the callsign CRC filter.

DJ2LS:
We could use the beacon as a heartbeat maybe. If we received a ping with a special crc, calculated by an empty Callsign or some wildcard character, we then answer with a beacon. But answering with a ping ack may be the cleanest way in my opinion. 
[...]

PR #158, opened 2022-04-10
WIP:Implement feature request #156.

kronenpj:
This initiates a PING upon receipt of an unaddressed PING.
TODO: Add configuration option to enable/disable feature.

There are some possible abuse paths with this, so we may want to consider rate limiting, incoming callsign format verification and other protections. As coded a maximum of 15 stations can respond over the course of 2.4 seconds. This may be a bit excessive so we may want to limit this to 5.

Thoughts?


DJ2LS:
I'm not sure, if we should do this in another way. 
An answer by every receiving station with a ping (+ ping ack) could cause too much traffic on a channel. 

Why not responding to a CQ with a single frame which contains Callsign and grid. This information could also be usable by other stations as well instead of direct addressing. 

We could then increase time between a CQ frame, wait for Responses then repeat two more times. By doing it like this we could possibly receive more stations 🤔

DJ2LS:
@kronenpj   I did a test with black code formatting -  but at least I had to revert the changes. I hope this didn't break anything here. I suggest we are going to continue with discussion here https://github.com/DJ2LS/FreeDATA/issues/156

kronenpj:
I'm going to close this and think about a better way to accomplish what I'm considering.

PR #166, opened 2022-04-15
Remove frequency_offset from received_ping.

kronenpj:
Per your earlier comment regarding removing frequency_offset until it's decided to fully remove or re-implement.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -267,9 +267,9 @@ class DATA():
                 structlog.get_logger("structlog").debug("PING RECEIVED....")
                 # = self.get_frequency_offset(freedv)
                 # we need to fix this later
-                frequency_offset = "0"
+                # frequency_offset = 0
                 #print("Freq-Offset: " + str(frequency_offset))
-                self.received_ping(bytes_out[:-2], frequency_offset)
+                self.received_ping(bytes_out[:-2])
                 
 
             # PING ACK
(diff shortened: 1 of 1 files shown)


DJ2LS:
Thanks for the cleanup! I suggest removing the entire Frequency Offset part!
Instead of changing the frequency we should consider doing this by an audio pitch and not by changing the VFO frequency of the radio. But this is not needed anymore as its meanwhile possible adjusting the freedv tuning range which solves most problems caused by an frequency Offset > +/- 50Hz

kronenpj:
Per your request. :)

tnc/data_handler.py
```diff
@@ -1503,7 +1486,7 @@ def received_ping(self, data_in:bytes):
         ping_frame[1:3] = static.DXCALLSIGN_CRC
         ping_frame[3:5] = static.MYCALLSIGN_CRC
         ping_frame[5:11] = static.MYGRID
-        ping_frame[11:13] = bytes([0,0]) # frequency_offset.to_bytes(2, byteorder='big', signed=True)
+        ping_frame[11:13] = bytes([0,0])
```
DJ2LS, line 1489: This line can be removed as we don't need this information anymore.

kronenpj:
I hadn't bothered to look up whether `bytearray` fills with zero. It does, so this can safely be removed.

DJ2LS (approved):

PR #167, opened 2022-04-15
Adding TNC log for undirected PING receipt.

kronenpj:
Restores a small chunk of code left out of prior merge. 
Correctly logs TNC's receipt of a valid PING packet but without destination call.


DJ2LS:
I'm not sure, if we should add support for undirected pings, when introducing #156 , as a ping should (theoretically) always be directed :thinking: 
What do you think about this?

kronenpj:
Oh, right. This is not needed. To enforce this idea, the UI should not allow the user to initiate a PING without populating the destination call field.

Issue #179, opened 2022-04-24
Initial ARQ open transition from connecting to connected edge case problem


kronenpj:
**Describe the bug**

**To Reproduce**
1. Connect two instances (first and second instance/station) via audio devices, omit one TX->RX path.
2. From (first) instance with good TX->RX path, initiate chat to second instance.
3. Observe output from `modem.py` on both sides.

**Expected behavior**
Second instance "gives up" after reasonable # of retries.

**Logfiles**
First Instance:
```txt
2022-04-24 09:30.32 [C2 ] Searching for libcodec2...
2022-04-24 09:30.32 [C2 ] Libcodec2 loaded         path=codec2/build_linux/src/libcodec2.so
2022-04-24 13:30:32 [info     ] [DMN] Starting TCP/IP socket   port=3001
2022-04-24 13:30:32 [info     ] [DMN] Starting FreeDATA Daemon author=DJ2LS version=0.3.3-alpha year=2022
2022-04-24 13:30:33 [debug    ] [SCK] Client connected         ip=127.0.0.1 port=58316
2022-04-24 13:30:50 [debug    ] [SCK] CMD                      command={'type': 'set', 'command': 'start_tnc', 'parameter': [{'mycall': 'N2KIQ-0', 'mygrid': 'EL98ip', 'rx_audio': '0', 'tx_audio': '0', 'radiocontrol': 'disabled', 'devicename': 'RIG_MODEL_DUMMY_NOVFO', 'deviceport': '/dev/ttyS0', 'pttprotocol': 'USB', 'pttport': '/dev/ttyS0', 'serialspeed': '9600', 'data_bits': '8', 'stop_bits': '1', 'handshake': 'None', 'rigctld_port': '4532', 'rigctld_ip': '127.0.0.1', 'enable_scatter': 'False', 'enable_fft': 'True', 'enable_fsk': 'False', 'low_bandwith_mode': 'True', 'tuning_range_fmin': '-50.0', 'tuning_range_fmax': '50.0', 'tx_audio_level': '1'}]}
2022-04-24 13:30:50 [debug    ] [DMN] TNC Startup Config : mycall value=N2KIQ-0
2022-04-24 13:30:50 [debug    ] [DMN] TNC Startup Config : mygrid value=EL98ip
2022-04-24 13:30:50 [debug    ] [DMN] TNC Startup Config : rx_audio value=0
[...]

PR #180, opened 2022-04-24
Initial attempt to fix issue #179.

kronenpj:
Made a single location where the ARQ state can transition from 'connecting' to 'connected'. Added a few checks of ARQ_SESSION_STATE where I thought they were missing.

Added sending of close session frame when the originating side times out. This can allow for the far-side to transition back to disconnected as well, which also stops the initial problem on non-updated nodes, if they receive the close session frame, without waiting for the 5-minute ARQ session timeout.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -1011,8 +1011,8 @@ class DATA():
             time.sleep(0.01)
             static.ARQ_SESSION_STATE = 'connecting'
 
-        if static.ARQ_SESSION:
-            static.ARQ_SESSION_STATE = 'connected'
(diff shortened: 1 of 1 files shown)


DJ2LS:
I didn't have a look at this yet. So I don't know how the changes will affect the problem. 
You could also add it to the v0.4.0 
PR if you want instead of creating a new one. But we need to be careful when working both at the same time :-) 
V0.4.0 introduces minor protocol changes regarding to CRCs. So it would be good if you could test this as well.

kronenpj:
Agreed, we've made changes in some of the same places. I didn't quite put 2+2 together this morning on your comment on wanting to fix this in v0.4.0. I'll re-do this on that branch and commit.

Issue #173, opened 2022-04-19
Session disconnect too short --> causes heartbeat toggle


DJ2LS:
Session disconnect too short --> causes heartbeat toggle.
We need to track down this problem as its causes a really bad tnc state.

DJ2LS:
We also need to avoid the toggle somehow in case of a disconnect frame loss

kronenpj:
Unfortunately this fix causes the receiving TNC to remain in BUSY / OPEN after experiencing the same RF / audio state which caused issue #179. In this state, the TNC does not properly respond to GUI commands until the session times out in ~12 minutes.  

I may start writing tests for the TNC state machine transitions. I need to understand the `ctest` approach first.

DJ2LS:
Thanks, @kronenpj for having a look!

On which side does the TNC not respond to commands? ISS or IRS ?

Regarding to the ctests, please also have a look at https://github.com/DJ2LS/FreeDATA/issues/102

kronenpj:
IRS. I implemented the TODO from this commit and it fixes this case as well.

DJ2LS:
mh, there seems to be another problem with connecting - actually we can't establish a ARQ connection anymore... 🤔

kronenpj:
This appears to be fixed with PR #183.

PR #184, opened 2022-05-02
Corrects issue #173 fix

kronenpj:
I didn't look carefully enough at `helpers.check_callsign` to see that it returns a list. The first element of the list is `True`/`False`.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -1140,7 +1140,8 @@ class DATA():
         Returns:
         """
         # Close the session if the DXCALLSIGN_CRC matches the station in static.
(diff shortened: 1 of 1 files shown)


DJ2LS (commented):
I would suggest we are doing it like here so we have a consistent checking within the entire application:


https://github.com/DJ2LS/FreeDATA/blob/c42be2e1d01a7bb83cc628f0687d310585f12a94/tnc/data_handler.py#L208

kronenpj:
Agreed. Made appropriate changes to patch.

DJ2LS:
I'm going to merge this now 👍

DJ2LS (approved):

Issue #102, opened 2022-01-04
Bidirectional piping instead of virtual audio devices for ctests


DJ2LS:
Bidirectional piping instead of virtual audio devices for ctests, so we can run the tests on GithubActions

Ideas:

- python audio dummy module
- mkfifo

drowe67:
Sending samples over UDP on localhost might work too.  Then `netcat` can be used to glue things together.

DJ2LS:
@drowe67 I decided not to update the actual ctests from pyAudio library to sounddevice library ( bot using portaudio ) within the latest version 0.4.0 . 
I'm not happy with running tests with virtual sound devices as this is not reliable enough for running the ctests. I'm thinking about named pipes or netcat for sending data like done with our simple stdio tests via pipes. For doing this we have to completely redesign them - so I don't want to spent too much time on the actual virtual audio device tests.

PR #185, opened 2022-05-04
Enhance tests

kronenpj:
Two new tests to catch recent coding errors and situations.


kronenpj:
Note that these would have caught my error which caused PR#184.

DJ2LS:
Great 🥳  you are working on the tests as well. I'm actually doing experiments with mkfifo so the full tnc can be tested - including the audio processing part, hopefully soon.

The idea is, completely redesigning the tests so we don't have to stay at the existing ones. They have been fine in a earlier TNC state, but now we need to adjust them.

My goal is, making tests with the protocol as well. 
- [ ] CQ 
- [ ] Ping
- [ ] Connect
- [ ] File transfer
....

kronenpj:
Am I correct in noticing that some of the tests in `FreeDATA` actually test the `codec2` code rather than `FreeDATA`'s use of that code? If so, I would advocate contributing them to the `codec2` repository and concentrate on testing `FreeDATA`-specific code.

DJ2LS:
> Am I correct in noticing that some of the tests in `FreeDATA` actually test the `codec2` code rather than `FreeDATA`'s use of that code? If so, I would advocate contributing them to the `codec2` repository and concentrate on testing `FreeDATA`-specific code.

hi @kronenpj , no, the tests here are only for FreeDATA and to see how the Python implemention/access of libcodec2 interacts with libcodec2, implemented in C. The tests here in Python are comparable to the tests done in codec2, so its easier detecting problems. We created them in a early phase of the project where not that much Python code existed for libcodec2. So its possibly some kind of proof of concept, as well.
But I want to redesign the overall tests ( and like you already started working on this as well ). So I agree, we might focus on protocol and tnc tests in future which are more FreeDATA related.

DJ2LS:
@kronenpj I think we should adjust the readme file as well. Maybe with a list of all tests we are doing and a kind of "test id" so we are know which tests we are already running and which areas are covered by them 🤔
I'm busy actually, so it may take some days until I can work here again.

kronenpj:
This PR may also help realize Issue #102.  The latest commit uses named pipes in `test_tnc.py`, with `test_tnc_IRS.py` and `test_tnc_ISS.py` helpers.

kronenpj:
I'm going to recommend that we merge this. These are pretty much done, fir as much as they do and I've started working on a new, more involved test. I dd notice that, while valid, `test_modem` doesn't actually do much. I expect I'll expand it later.

DJ2LS:
Agree. 
How's the bidirectional test working now? So we can write a small todo list for further improvements here before merging and reopening it in future

kronenpj:
My thought is to create a more-or-less generic test framework for these, so that it's easier to create tests without having to dig into the details every time. I'm getting there, but it won't be ready and I'd like to be able to use the current tests with new TNC code. It's already a pretty big commit and I prefer smaller ones that are more readable.
I need to find a way to capture what the receive side for comparison to what's sent, data, frame type, callsigns, etc..
It's almost looking to me like the 'lost' first frame isn't actually lost, it's almost like it's sitting in a buffer waiting for more information before being processed. I'm looking into it.

DJ2LS:
Yes I agree with you. Just merge if you are ready! 
My thoughts with the todo list have been, that if we are going to work again on this, we can continue directly without having to look again too much at what's working and what's not. Sorry if I wrote misleading.

PR #187, opened 2022-05-15
Refactor TNC modules

kronenpj:
A set of changes to improve the 'pythonness' of the TNC modules. I've manually verified these changes did not affect functionality. More changes are likely on the way but these are good as-is. I'll make sure everything works before any pushing any future commits to this branch.

### DIFF
--- a/tnc/audio.py
+++ b/tnc/audio.py
@@ -1,9 +1,10 @@
 
-import json
-import sys
-import multiprocessing
-import sounddevice as sd
 import atexit
+import json
+import multiprocessing
+import sys
+
+import sounddevice as sd
 
 atexit.register(sd._terminate)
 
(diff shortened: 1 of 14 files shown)


DJ2LS:
@kronenpj this is awesome work! 🙂 We should merge this as soon as possible.

This makes me also thinking about a code style guideline.

kronenpj:
It's ready to be merged today, if you'd like. I'm currently pondering how to create tests for `modem.py` and whether `PyAudio` can be used with `stdio`. It would help me think more clearly if both this and #185 were merged but they don't depend on each other.

I've tried to use PEP8 (via the `black` formatter and `isort`) as primary guidelines. The use of an integrated development environment such as VSCode or PyCharm helps quite a bit as well.

DJ2LS:
Okay, then lets merge them! 

I did a change to data_handler.py which uses named pipes for data exchange without PyAudio. But this needs some more work.

Issue #207, opened 2022-06-05
Final data packet in burst does not retry on NACK.


kronenpj:
**Describe the bug**
The chat / file transfer OTA protocol doesn't resend the last frame if a NACK is received. Instead the routine drops into the transmission failed state and closes the connection.

**To Reproduce**
1. Craft a data frame that is larger than one frame (>~560 bytes)
2. Send over unreliable or delayed channel like FIFOs.
3. Wait for file transfer to complete / run new `test_data_channel` test.
4. Observe debug output / notice large chat or file transfer failed.

**Expected behavior**
Final frame is properly transmitted when NACK is received.

**Logfiles**
```
2022-06-05 17:56:23 [info     ] [TNC] ARQ | DATA | TX | [AA2BB-0]>> <<[ZZ9YY-0] attempt=1/5
2022-06-05 17:56:23 [info     ] S1 TX:                         TX=225
2022-06-05 17:56:26 [info     ] [TNC] ARQ | DATA | TX | [AA2BB-0]>> <<[ZZ9YY-0] attempt=2/5
2022-06-05 17:56:26 [info     ] S1 TX:                         TX=225
2022-06-05 17:56:26 [info     ] S2 RX:                         RX=225
2022-06-05 17:56:26 [info     ] [TNC] ARQ | DATA | RX | [ZZ9YY-0]>> <<[AA2BB-0] bandwidth=wide
2022-06-05 17:56:26 [info     ] S2 TX:                         TX=228
2022-06-05 17:56:26 [info     ] [TNC] ARQ | DATA | RX | [ZZ9YY-0]>>|<<[AA2BB-0] bandwidth=wide snr=9.0
2022-06-05 17:56:27 [warning  ] [TNC] Frame timeout            attempt=0 max_attempts=50 speed_level=2
2022-06-05 17:56:27 [info     ] S2 TX:                         TX=64
2022-06-05 17:56:27 [info     ] S1 RX:                         RX=228
2022-06-05 17:56:27 [info     ] [TNC] ARQ | DATA | TX | [AA2BB-0]>>|<<[ZZ9YY-0] snr=10.1
2022-06-05 17:56:27 [info     ] [TNC] | TX | DATACHANNEL       Bytes=1530 mode=10
2022-06-05 17:56:27 [info     ] [TNC] ARQ | TX | FRAMES        fpb=1 mode=10 retry=0
2022-06-05 17:56:27 [info     ] S1 TX:                         TX=10
[...]

PR #208, opened 2022-06-05
Fix for issue #207.

kronenpj:
This was found during test development. The fix is almost functionally equivalent to the original, but allows for the final frame to be re-sent the same as other frames.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -865,11 +865,7 @@ class DATA:
         bufferposition = bufferposition_end = 0
 
         # Iterate through data_out buffer
-        while (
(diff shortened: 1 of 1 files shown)


DJ2LS (approved):
Checked from couch. Should work in my opinion. @kronenpj does this also work when running the tnc?

kronenpj:
Haven't tested on the soundcard connected systems, but I also don't know if I can stimulate the bug from the GUI. Let me do that now...

DJ2LS:
I'm testing these things OTC/over the cable and switching the frequency manually in case I want to simulate a lost frame. But I can run a short test in two days after work, if you want.

PR #216, opened 2022-06-17
Add FRAME_TYPE enum to static

kronenpj:
Add a FRAME_TYPE enumeration to use primarily for logging but going forward in the creation of frames rather than plain integers.


tnc/static.py
```diff
@@ -123,3 +124,43 @@
...
+    FR_REPEAT = 62
+    FR_NACK = 63
+    BURST_NACK = 64
+    CQ = 200
+    QRV = 201
+    PING = 210
+    PING_ACK = 211
+    ARQ_OPEN = 221
+    ARQ_HB = 222
+    ARQ_CLOSE = 223
```
DJ2LS, line 159: Let's call this ARQ_SESSION_xxx so it's clear where the frame belongs to.

DJ2LS:
Great idea, @kronenpj ! 
We should add this to the wiki as well as a table or something like that 🤔

DJ2LS (approved):

PR #221, opened 2022-06-21
Use FRAME_TYPE enum instead of plain int.

kronenpj:
Move to using the statoc.FRAME_TYPE enumeration instead of integers for frame types in data_handler.py.
Update FRAME_TYPE enum for missing items.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -347,13 +347,6 @@ class DATA:
                 self.log.debug("[TNC] ARQ arq_received_channel_is_open")
                 self.arq_received_channel_is_open(bytes_out[:-2])
 
-            # ARQ MANUAL MODE TRANSMISSION
-            elif (
-                FR_TYPE.ARQ_MANUAL_01.value <= frametype <= FR_TYPE.ARQ_MANUAL_11.value
-            ):
-                self.log.debug("[TNC] ARQ manual mode")
-                self.arq_received_data_channel_opener(bytes_out[:-2])
-
             # ARQ STOP TRANSMISSION
             elif frametype == FR_TYPE.ARQ_STOP.value:
                 self.log.debug("[TNC] ARQ received stop transmission")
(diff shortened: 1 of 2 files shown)


tnc/data_handler.py
```diff
@@ -261,92 +272,100 @@ def process_data(self, bytes_out, freedv, bytes_per_frame: int) -> None:
...
             # ARQ CHANNEL IS OPENED
-            elif frametype in [226, 228]:
+            elif frametype in [
+                FR_TYPE.ARQ_DC_OPEN_ACK_W.value,
+                FR_TYPE.ARQ_DC_OPEN_ACK_N.value,
+            ]:
                 self.log.debug("[TNC] ARQ arq_received_channel_is_open")
                 self.arq_received_channel_is_open(bytes_out[:-2])
 
             # ARQ MANUAL MODE TRANSMISSION
```
DJ2LS, line 350: In my opinion we can remove the manual mode selection, as a automatic selection fits best, specially under really bad HF conditions. This feature works well in development environments but not under real circumstances

tnc/data_handler.py
```diff
@@ -413,7 +432,7 @@ def send_data_to_socket_queue(self, /, **jsondata):
     def send_burst_ack_frame(self, snr) -> None:
         """Build and send ACK frame for burst DATA frame"""
         ack_frame = bytearray(14)
-        ack_frame[:1] = bytes([60])
+        ack_frame[:1] = bytes([FR_TYPE.ACK.value])
```
DJ2LS, line 435: This one should be named BURST_ACK so it's homogeneous to the other names.

tnc/static.py
```diff
@@ -161,6 +148,9 @@ class FRAME_TYPE(Enum):
     ARQ_DC_OPEN_ACK_W = 226
     ARQ_DC_OPEN_N = 227
     ARQ_DC_OPEN_ACK_N = 228
+    ARQ_MANUAL_01 = 230
+    # ...
```
DJ2LS, line 152: Same thought here with removing like above 🤔

DJ2LS (approved):
Approved by checking from mobile phone at work. Wasn't able testing this by running the application.

tnc/static.py
kronenpj: The latest push removes the manual mode code.

tnc/data_handler.py
kronenpj: Done.

DJ2LS (approved):

PR #222, opened 2022-06-21
Remove allmodes (255) option from TNC.

kronenpj:
First attempt.

I don't know if/how we want to handle removing this from the GUI piece but the TNC doesn't pay much attention to the `mode` value the GUI passes anyway. I didn't completely take out the code in `data_handler` around line `2410` so it will listen to all modes unless one is specified. Also, the `manual mode` code is still around for the manual mode, so that's still a possibility.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -25,6 +25,7 @@ import static
 import structlog
 import ujson as json
 from exceptions import NoCallsign
+from codec2 import FREEDV_MODE
 
 TESTMODE = False
 
@@ -84,17 +85,15 @@ class DATA:
 
         # List of codec2 modes to use in "low bandwidth" mode.
         self.mode_list_low_bw = [
-            codec2.FREEDV_MODE.datac0.value,
-            codec2.FREEDV_MODE.datac3.value,
+            FREEDV_MODE.datac3.value,
         ]
         # List for time to wait for corresponding mode in seconds
         self.time_list_low_bw = [3, 7]
 
         # List of codec2 modes to use in "high bandwidth" mode.
         self.mode_list_high_bw = [
-            codec2.FREEDV_MODE.datac0.value,
-            codec2.FREEDV_MODE.datac3.value,
-            codec2.FREEDV_MODE.datac1.value,
+            FREEDV_MODE.datac3.value,
+            FREEDV_MODE.datac1.value,
         ]
         # List for time to wait for corresponding mode in seconds
         self.time_list_high_bw = [3, 7, 8, 30]
@@ -360,7 +359,7 @@ class DATA:
     def enqueue_frame_for_tx(
         self,
         frame_to_tx: bytearray,
-        c2_mode=codec2.FREEDV_MODE.datac0.value,
+        c2_mode=FREEDV_MODE.datac0.value,
         copies=1,
         repeat_delay=0,
     ) -> None:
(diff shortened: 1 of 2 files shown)


tnc/data_handler.py
```diff
@@ -876,39 +876,8 @@ def arq_transmit(self, data_out: bytes, mode: int, n_frames_per_burst: int):
...
-                    #    self.speed_level = max(self.speed_level - 1, 0)
-
-                    # if self.tx_n_retry_of_burst <= 1:
-                    #    self.speed_level += 1
-                    # self.speed_level = max(self.speed_level + 1, len(self.mode_list) - 1)
-
-                    # Bound speed level to:
-                    # - minimum of either the speed or the length of mode list - 1
-                    # - maximum of either the speed or zero
-                    self.speed_level = min(self.speed_level, len(self.mode_list) - 1)
```
DJ2LS, line 901: This and the lines following shouldn't be removed, I think.

For min and max functions I used np.clip, so I suggest we are using the same functions application wide. There has also be a bug with min/max in another area of the code. So we should test this, I think. I can do this as well the next days as I want to see the possible differences between np.clip and min/max 🤔

tnc/data_handler.py
```diff
@@ -876,39 +876,8 @@ def arq_transmit(self, data_out: bytes, mode: int, n_frames_per_burst: int):
         while not self.data_frame_ack_received and static.ARQ_STATE:
             # we have TX_N_MAX_RETRIES_PER_BURST attempts for sending a burst
             for self.tx_n_retry_of_burst in range(TX_N_MAX_RETRIES_PER_BURST):
-                # AUTO MODE SELECTION
-                # mode 255 == AUTO MODE
-                # force usage of selected mode
-                if mode != 255:
```
DJ2LS, line 882: This is the area which can be removed.

DJ2LS (changes requested):

tnc/data_handler.py
kronenpj: Restored that section. 

While I don't necessarily object to using the same function application-wide, `np.clip` is really meant for NumPY arrays, not single arrays. I can make the change as I'm sure it will work, but if we need to provide `np.clip` a list and retrieve element 0 from the returned list I would suggest we use min/max instead.

DJ2LS: Ah yes, you're right. We used np.clip when processing audio data I think. I just want to avoid a possible bug. So it's good if we are talking about this 👍 
I,ll resolve this conversation.

DJ2LS (approved):

kronenpj:
When I converted the high/low speed lists from integers to enumerations, I started wondering if including `datac0` in the list of data modes for high/low bandwidth data transfers was really the intent. The failure of the long message variant in `test_chat_text` shows that it may not be or at least that it doesn't work properly. Because of the "packet delay" in the named pipes, the tests exercise the speed changes and one that falls down to `datac0` consistently corrupts the checksum:

`2022-06-24 12:18:18 [warning  ] [TNC] ARQ | RX | DATA FRAME NOT SUCCESSFULLY RECEIVED! e=wrong crc expected=b'\x88\xdf\xaa\x80' overflows=[0, 0, 0, 0, 0] received=b'\xdb\xa8S\xe4'`

Removing `datac0` from the list clears up the test failure. If we want to include `datac0` in the list, we'll need to dig further into why the checksum is being corrupted. Thoughts?

DJ2LS:
My intention when including datac0 was, that it is the mode with the lowest snr operstion level. But we can remove it in my opinion as it causes several possible problems. For the future another data mode for lower snr is planned.

kronenpj:
Understood. I'll remove it from the list in this PR since it relates to speed changing.

DJ2LS (approved):
And again: your work is great🙂 This is a nice improvement and should make it easier reading the code 💪🏻

Issue #194, opened 2022-05-18
New PyInstaller (v5.1) breaks build


kronenpj:
The github action to build our release fails (on the soundmodem package) with PyInstaller 5.1.

**To Reproduce**
1. Change 'pyinstaller<5.1' to 'pyinstaller'
2. Re-run the github action.
3. Observe failure.

**Expected behavior**
Build succeeds.

**Logfiles**
One example.
https://github.com/DJ2LS/FreeDATA/runs/6497911314?check_suite_focus=true#step:4:100

Workaround committed to main: Limit pyinstaller to versions < 5.1.

DJ2LS:
test run with pyinstaller 5.1 and python 3.9.7 succeeded on Ubuntu 20.04 🤔

kronenpj:
Interesting, all I can say is that it was failing to build the executables when I was trying to release alpha4-7. I was able to trace it back to 5.1 but I fully believe the real problem has been fixed and 5.1 is working again.

DJ2LS:
We could test this carefully with the next release. On success: 👍 , if it fails: Fall back to older Vera hin which will be released as a hot fix

PR #224, opened 2022-06-21
Allow pyinstaller version to float again.

kronenpj:
Forgot to update these when previous release was created.  This will test whether issue #194 can be closed as OBE.

### DIFF
--- a/.github/workflows/build-project-linux.yml
+++ b/.github/workflows/build-project-linux.yml
@@ -35,7 +35,7 @@ jobs:
           pip3 install pyaudio
           pip3 install psutil
           pip3 install crcengine
-          pip3 install 'pyinstaller<5.1'
+          pip3 install pyinstaller
           pip3 install ujson
           pip3 install pyserial
           pip3 install numpy
--- a/.github/workflows/build-project-mac.yml
+++ b/.github/workflows/build-project-mac.yml
@@ -33,7 +33,7 @@ jobs:
           python -m pip install --upgrade pip
           pip3 install psutil
           pip3 install crcengine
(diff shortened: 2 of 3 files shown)


kronenpj:
My experiment didn't workas I expected and I accidentally committed two more tags to this repository. In any event this may or may not work... More testing required.

.github/workflows/build-project-linux.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           pip3 install pyaudio
           pip3 install psutil
           pip3 install crcengine
-          pip3 install 'pyinstaller<5.1'
```
DJ2LS, line 38: We should keep this in mind, maybe adding a comment with this and a small description. There will be some version problems possibly somewhen in future ( not only with pyinstaller). So we already have a working solution of limiting specific module versions

DJ2LS (approved):

.github/workflows/build-project-linux.yml
kronenpj: It is common practice to anchor / freeze the dependency versions to ones known to work. While we're in Alpha / Beta it's probably acceptable to float, but once we hit 'production' the versions should be frozen to get consistent builds.

DJ2LS: Yes, good point and I agree! We should talk somewhen the next days about a roadmap and a possible first beta release.

kronenpj: We may want to migrate to using [`poetry`](https://python-poetry.org/) over `pip` just because it handles that more gracefully, in my opinion. Project for a future time.

PR #229, opened 2022-06-24
Add environment creation scripts.

kronenpj:
Suggest these scripts to setup python and node/npm environments locally, without needing `sudo`. This is how I've been installing and developing FreeDATA since the beginning. The [wiki](https://wiki.freedata.app) instructions still work of course.


kronenpj:
I think I should add a batch script as well. Though it's difficult for me to test as I have limited access to Windows machines.

DJ2LS:
Oh, I didn't expect such a fast PR. I just "saved" my thoughts regarding to setup scripts 🙂

Three further thoughts:
- One, combined bash script for installing npm and python modules 
- Script should be placed inside "tools" and should be executed from there, in my opinion.
- Another approach could be, one setup script which includes git clone and also npm and python dependency installation 🤔

kronenpj:
I prefer to keep the scripts single-purpose if possible. I could create a script which calls each of the others. Scripting a `git clone` introduces a chicken-and-egg problem since the cloning script would be inside the repository. 

I'll move them to the tools directory once I get them working from there.

tools/create_python_env.sh
```diff
@@ -0,0 +1,48 @@
...
+# Activate the virtual environment, if needed
+if [ -z "${VIRTUAL_ENV}" ] || [ "${VIRTUAL_ENV}" != "${VENVDIR}" ]; then
+    source "${VENVDIR}/activate"
+fi
+
+# Cease using ${CHOSEN} as the interpreter we want now is in our path.
+
+# Install packages
+python3 -m pip install -U pip wheel
+python3 -m pip install -r requirements.txt
```
DJ2LS, line 45: This is also a part we could use for GitHub Actions so we have one homogeneous data source

DJ2LS (approved):
This is a good start, I think. We could also update the wiki as soon as they are running stable and fit for our needs. These scripts could avoid some bugs caused by missing packets not installed by some users.

DJ2LS (commented):
Regarding to the requirements.txt there are some side dependencies it seems. I'm wondering if we can mark them by some notes 🤔

kronenpj:
Looks like I forgot a commit to weed out dependencies. :)

PR #231, opened 2022-06-25
Create get_audio_devices list from set.

### LINKED ISSUE #174: sounddevice present multiple times in the dropdown list
The speaker/microphone sounddevice is always present multiple times in the respective dropdown list.
It doesn't seem to make a difference which one you choose, but it is irritating.

Problem occured on Windows 10 and Windows 11:

image

kronenpj:
This eliminates duplicates without a lot of gyrations.

We should consider a release from this branch to see if this fix addresses @DK5SM's issue in #174 unless he can build from source to verify.

### DIFF
--- a/tnc/audio.py
+++ b/tnc/audio.py
@@ -35,7 +35,7 @@ def get_audio_devices():
         proc.join()
 
         # The use of set forces the list to contain only unique entries.
-        return list(set(proxy_input_devices)), list(set(proxy_output_devices))
+        return list(proxy_input_devices), list(proxy_output_devices)
 
 
 def fetch_audio_devices(input_devices, output_devices):
@@ -50,7 +50,10 @@ def fetch_audio_devices(input_devices, output_devices):
 
     """
     devices = sd.query_devices(device=None, kind=None)
-    for index, device in enumerate(devices):
+    input_devs = set()
+    output_devs = set()
+
+    for device in devices:
         # Use a try/except block because Windows doesn't have an audio device range
         try:
             name = device["name"]
@@ -65,6 +68,11 @@ def fetch_audio_devices(input_devices, output_devices):
             name = ""
 
         if max_input_channels > 0:
-            input_devices.append({"id": index, "name": name})
+            input_devs.add(name)
         if max_output_channels > 0:
-            output_devices.append({"id": index, "name": name})
+            output_devs.add(name)
+
+    for index, item in enumerate(input_devs):
+        input_devices.append({"id": index, "name": item})
+    for index, item in enumerate(output_devs):
+        output_devices.append({"id": index, "name": item})
--- a/tnc/audio.py
+++ b/tnc/audio.py
@@ -34,7 +34,6 @@ def get_audio_devices():
         proc.start()
         proc.join()
 
-        # The use of set forces the list to contain only unique entries.
         return list(proxy_input_devices), list(proxy_output_devices)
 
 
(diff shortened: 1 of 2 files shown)


kronenpj:
Unfortunately this fix emits:
`2022-06-25 20:04:20 [error    ] [DMN] update_audio_devices: Exception gathering audio devices: e=TypeError("unhashable type: 'dict'")` so a method heavier in gyration will need to be used.

DJ2LS (approved):
Interesting approach. This could work I think. 
We could publish this now as a release so we are staying at one Update branch, So we are not confusing testers because of different release origins. This might become easier when moving to a first beta release. 
Early testers can use alpha releases then. All others should stay at beta or stable.

DJ2LS:
@kronenpj I'm not sure, but can we merge this? Long pause

DJ2LS (approved):

dk5sm:
Paul,

I am sorry to say, but this seems not to work either (started FreeDATA.exe from /win-unpacked folder):
image

DJ2LS:
I'm not sure if the build process is working 100% as expected, but it might be possible this is a gui problem. So we should do some filtering on gui side as well. 

As soon as @dk5sm posted the select part of the html code, we can determine how we can do filtering.

dk5sm:
image
image

DJ2LS:
https://github.com/DJ2LS/FreeDATA/blob/690f400a81d975d470c25f6cd89b145bdf7d7c6c/gui/preload-main.js#L1380


Area of interest

DJ2LS:
I suggest we should also add a crc check for determining if it's really a unique audio device like done with serial devices. So we can also check if it's a  client specific behavior


```

          for port, desc, hwid in ports:
                    # calculate hex of hwid if we have unique names
                    crc_hwid = self.crc_algorithm(bytes(hwid, encoding="utf-8"))
                    crc_hwid = crc_hwid.to_bytes(2, byteorder="big")
                    crc_hwid = crc_hwid.hex()
                    description = f"{desc} [{crc_hwid}]"
                    serial_devices.append(
                        {"port": str(port), "description": str(description)}
                    )

```

kronenpj:
Gah! The code checkout portion of the build script tied the build to the `main` branch rather than the one that caused the action to start. Output from this run will have the fixes from this pull request:
https://github.com/DJ2LS/FreeDATA/actions/runs/3086117718

tnc/audio.py
```diff
@@ -49,21 +64,33 @@ def fetch_audio_devices(input_devices, output_devices):
...
+            continue
         except Exception as err:
             print(err)
             max_input_channels = 0
             max_output_channels = 0
             name = ""
 
         if max_input_channels > 0:
-            input_devices.append({"id": index, "name": name})
+            input_devs.add(device_crc(name))
```
DJ2LS, line 89: We should also include additional information like channels, supported sample rate and so on ( maybe all information which is stored in "devices"), so we can be sure we have a unique checksum, even if the names are equal ( happens with Icom ic705 for example. Same usb device name but different settings)

DJ2LS (approved):

tnc/audio.py
kronenpj: Done. All data in the `device` structure is included in the checksum.

DJ2LS (approved):

DJ2LS:
Nice, @kronenpj ! Let's see if these steps help us in debugging.

We can merge it in my opinion, then let's see if the crc is unique enough.
I'm really interested what's the origin of the topics problem 🤔

Issue #226, opened 2022-06-23
Setup bash script for installing dependencies


DJ2LS:
A updater and setup script would make it easier installing FreeDATA. Specially for user running FreeDATA from source.

PR #232, opened 2022-06-25
Add useful exclusions

kronenpj:
I keep stashing these but with the scripts from #226 this will become useful to more than just me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,18 @@ tnc/codec2
 
 package-lock.json
 .DS_Store
+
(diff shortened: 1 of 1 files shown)


DJ2LS (approved):
pycharm also produces some files we don't want to see within the repository. I need to update them somewhen later, as well

kronenpj:
Agreed. I Included `.idea` and `*.iml` which are generated by IDEA with the PyCharm plugin. It's been a while since I switched to Visual Studo Code so things may have changes.

PR #234, opened 2022-07-02
Refactor N2KIQ 202207

kronenpj:
Re-thinking the if-elif-else tree in `process_data` and similarly some TNC command processing.

### DIFF
--- a/tnc/data_handler.py
+++ b/tnc/data_handler.py
@@ -481,19 +481,17 @@ class DATA:
 
         Returns:
         """
+        # We've arrived here from process_data which already checked that the frame
+        # is intended for this station.
         data_in = bytes(data_in)
 
         # get received crc for different mycall ssids
         self.received_mycall_crc = data_in[2:5]
 
         # check if callsign ssid override
-        _valid, mycallsign = helpers.check_callsign(
+        _, mycallsign = helpers.check_callsign(
             self.mycallsign, self.received_mycall_crc
         )
-        if not _valid:
-            # ARQ data packet not for me.
-            self.arq_cleanup()
-            return
 
         # only process data if we are in ARQ and BUSY state else return to quit
         if not static.ARQ_STATE and static.TNC_STATE != "BUSY":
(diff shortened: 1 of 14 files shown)


tnc/data_handler.py
```diff
@@ -56,7 +53,7 @@ def __init__(self) -> None:
         self.transmission_uuid = ""
 
         # Received my callsign crc if we received a crc for another ssid
-        self.received_mycall_crc = b""
+        self.received_mycall_crc = b""  # Does this need to be a class variable?
```
DJ2LS, line 56: I'm not sure about its usage right now. My intention was, using class wide variables for callsigns instead of using static.XXXX while an ongoing arq transmission. But might it be possible, this variable is unused?

tnc/data_handler.py
```diff
@@ -126,6 +123,57 @@ def __init__(self) -> None:
 
         self.transmission_timeout = 360  # transmission timeout in seconds
 
+        # Dictionary of functions and log messages used in process_data
+        # instead of a long series of if-elif-else statements.
+        self.rx_dispatcher = {
```
DJ2LS, line 128: Nice! But we need to be carefully with older Python Versions. I'm not sure if this is working for older versions like 3.6 for example? We shouldn't focus that much on outdated software, but a lot of LTS operating systems are still using 3.6, I think.

DJ2LS (changes requested):

tnc/data_handler.py
kronenpj: It's set in three places:
`__init__` on line 56,
`arq_data_received` on line 448,
`arq_cleanup` on line 2288

It's never used anywhere, only set.

tnc/data_handler.py
kronenpj: I'll specifically check on Python 3.6 but it's not a new feature - everything in Python is an object, including functions.

tnc/data_handler.py
DJ2LS: Then we can remove it, I think 🙂

tnc/data_handler.py
DJ2LS: Yes, no worries. I'm just not exactly sure, so better checking twice. But if you're saying it's okay, then we can merge it 👍

DJ2LS (approved):

tnc/data_handler.py
kronenpj: Nope, there are a few problems with Python 3.6...  Fixing... :)

tnc/data_handler.py
kronenpj: Removed...

tnc/data_handler.py
kronenpj: Fixed.

Issue #240, opened 2022-09-04
Filetransfer broken


DJ2LS:
2022-09-04 17:28:58 [debug    ] [TNC] send_data_to_socket_queue: jsondata={'freedata': 'tnc-message', 'arq': 'transmission', 'status': 'transmitting', 'uuid': '2acd0fa4-db7e-4cb8-94ac-eecab8063f56', 'percent': 0, 'bytesperminute': 0}
Exception in thread worker thread transmit:
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1009, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 946, in run
    self._target(*self._args, **self._kwargs)
  File "/home/dj2ls/PycharmProjects/FreeDATA/tnc/data_handler.py", line 232, in worker_transmit
    self.open_dc_and_transmit(data[1], data[2], data[3], data[4], data[5])
  File "/home/dj2ls/PycharmProjects/FreeDATA/tnc/data_handler.py", line 1491, in open_dc_and_transmit
    self.arq_transmit(data_out, mode, n_frames_per_burst)
  File "/home/dj2ls/PycharmProjects/FreeDATA/tnc/data_handler.py", line 808, in arq_transmit
    mode=FREEDV_MODE(mode).name,
  File "/usr/lib/python3.10/enum.py", line 385, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib/python3.10/enum.py", line 710, in __new__
    raise ve_exc
ValueError: 255 is not a valid FREEDV_MODE

PR #241, opened 2022-09-04
hotfix for #240 and failing GitHub action

DJ2LS:
hotfix for #240 which caused a failing file transfer. As its just a hotfix, we might improve this later. 
Also hopefully fixed a part of our new Github Action for arm - builds. Arm builds still not integrated in a working way into the new build, but as a first step we want a working build process.


DJ2LS (commented):
GitHub Action still broken for arm. Seems, there have been some changes outside FreeDATA/possibly codec2, which affect arm build. This should be fixed in another PR

PR #245, opened 2022-09-06
move pyinstaller to nuitka

DJ2LS:
Nutika transpiler has been released in version > 1.0 so a stable version can be expected.
Furthermore build process is simplified and we can expect a potential cpu usage improvement.


DJ2LS:
Also fixed in this PR: arm build process 💪🏻

Nevertheless, the new nuitka builds need to be tested on all platforms. 

Also the moment of a new release is exciting...not sure if a new release would work 🤔 only one way figuring this out...

PR #253, opened 2022-09-20
WIP: new rigctld control interface

DJ2LS:
We need to drop hamlib integration and rigctl support for moving to a better working rigctld integration.

- [x] Update Control Interface
- [x] Start / Stop Rigctld
- [x] Search for Rigctld executable

https://user-images.githubusercontent.com/75909252/191235895-2b9979bb-d4fa-4b17-8193-7820ce93d0f3.mp4


DJ2LS:
@kronenpj I'd like to merge this. Could you do me the favor and doing some review? This feature is working, but might need some more improvements for sure. So we should test this with real users as well, so publishing this as part of an alpha would be great before moving to a beta as this is a minor gui change. 
Hamlib support in general is dropped and only network based hamlib/rigctld support now available.
The TNC itself supports other control options as well, but we can drop them from time to time as well I think.

kronenpj (approved):
This looks good to me. The JS and HTML changes were mind-numbing but looked OK.

DJ2LS:
Thanks, @kronenpj . The GUI related parts need some optimization somewhen....Its a big pain I guess.

I'm doing some small tests here with 32bit building step for possibly adding 32bit support for windows which has been a issue on mailing list. Its at least a chance for fixing this.

PR #254, opened 2022-09-20
WIP: add config file support

DJ2LS:
We want config files for easier starting the tnc from CLI

- [x] write config to config file from daemon
- [x] read config when running tnc
- [x] update documentation https://wiki.freedata.app/en/usage/tnc

### DIFF
--- a/tnc/config.py
+++ b/tnc/config.py
@@ -1,0 +1,97 @@
+import configparser
+import structlog
+
+class CONFIG:
+    """
+    CONFIG class for handling with config files
+
+    """
+
+    def __init__(self):
+        # set up logger
+        self.log = structlog.get_logger("CONFIG")
+
+        # init configparser
+        self.config = configparser.ConfigParser(inline_comment_prefixes="#", allow_no_value=True)
+        self.config_name = "config.ini"
+
+        self.log.info("[CFG] logfile init", file=self.config_name)
+
+        # check if log file exists
+        self.config_exists()
+
+    def config_exists(self):
+        """
+        check if config file exists
+        """
+        try:
+            return bool(self.config.read(self.config_name, None))
+        except Exception as configerror:
+            self.log.error("[CFG] logfile init error", e=configerror)
+            return False
+
+    def write_config(self, section: str, key: str, value):
+        """
+        write values to config
+        """
+
+    def write_entire_config(self, data):
+        """
+        write entire config
+        """
+        #self.config['DEFAULT'] = {'':''}
+
+        self.config['NETWORK'] = {'#Network settings': None,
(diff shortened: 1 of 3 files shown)


DJ2LS:
example config stored in config.ini 

```
[NETWORK]
#network settings
tncport = 0

[STATION]
#station settings
mycall = DJ2LS-4
mygrid = JN48ea

[AUDIO]
#audio settings
rx = 3
tx = 3
txaudiolevel = 116

[RADIO]
#radio settings
radiocontrol = disabled
devicename = RIG_MODEL_IC705
deviceport = /dev/ttyACM0
serialspeed = 9600
pttprotocol = USB
pttport = /dev/ttyACM1
data_bits = 8
stop_bits = 1
handshake = None
rigctld_ip = 127.0.0.1
rigctld_port = 4532

[TNC]
#tnc settings
scatter = False
fft = True
narrowband = False
fmin = -150.0
fmax = 100.0
qrv = True
rxbuffersize = 16
```

kronenpj:
A couple of observations:
* The configuration file completely overrides settings supplied on the command line. This is probably not the behaviour expected. If I've taken the effort to supply a command line option to change something, it's probably more "important" than the one in the configuration file. This might be addressed by removing the `else` clause after reading the config file.
* Two chunks of code are duplicated between `daemon.py` and `main.py`. Could those be rolled into a separate file?

DJ2LS:
Yes, thats a good point. But if we are not going to separate both use-cases, config will be overwritten by default settings provided by argparse. I'll think how we could solve this.

kronenpj:
Think of it as a hierarchy. Consider a scheme like this, from lowest to highest priority (importance):
1. Configuration file contents, which are overwritten by
2. ArgParse Defaults for any missing items, which are overwritten by
3. Command line options

We can then give the option of saving the current configuration in the file. Thoughts?

DJ2LS:
> We can then give the option of saving the current configuration in the file. Thoughts?
The daemon actually writes the config as soon as the tnc will be started. So its easier creating an initial config file. 

Good idea with the hierarchy. This also a thing we should add to the wiki

DJ2LS:
mh it seems argparse is always falling back to default values if not set. So it seems, we have to consider if dropping "default" values in argparse and using defaults from static file instead. Or am I on the wrong way?

DJ2LS:
I'm trying now an approach with a lot of if - checks.

DJ2LS:
@kronenpj It seems the hierarchy feature is a harder nut than expected because of argparse. What do you think, can we merge this and working later on config file hierarchy? We should then open an issue for this.

kronenpj (approved):

PR #258, opened 2022-10-05
FIX: audio device list with unique entities

DJ2LS:
Audio device list "ID" item was wrong. This fix is another attempt so we have unique device names and a device can occur online once in device list


DJ2LS:
@kronenpj I changed logging for better readability and fixed the problem with device IDs have been corrupted. Maybe this is helping us by finding the problem with doubled audio devices on Windows.

PR #276, opened 2022-11-07
python3.7 compatibility

DJ2LS:
For raspberry pi we need python3.7 compatibility.

and some other interesting fixes.....


DJ2LS (commented):
seems to be working for me. Lets merge.

PR #277, opened 2022-11-08
reduce cpu load

DJ2LS:
- [ ] move rigctld check to a socket check to reduce cpu load specially on windows
- [ ] introduced a splash screen


DJ2LS (commented):
additional changes to gui and fixed issues

PR #299, opened 2022-12-03
fix ssid and response handling

DJ2LS:
- [x] dont respond to connection requests if we don't want them
- [x] override tnc ssid so we can use the tnc from different apps with a corresponding ssid
    - [x] raw data
        - [x] opening file session
        - [x] received file session
        - [x] received file
    - [x] connection / session
        - [x] open connection
        - [x] received connection 
        - [x] sending heartbeat
        - [x] received heartbeat
        - [x] opening file session
        - [x] received file session
        - [x] received file
    - [x] ping
        - [x] call
        - [x] receive
        - [x] received ACK 
    - [ ] cq
        - [ ] call - override not yet needed because of broadcast
        - [ ] receive - override not yet needed because of broadcast
        - [ ] received QRV - override not yet needed because of broadcast
- [x] tools for remote control via ssh
    - [x] send file
    - [x] network listener
        - [x] receive file
        - [x] transmission success
 - [x] set audio device by name


DJ2LS:
@dk5sm I'm doing some work on callsign ssids so it should be possible, using one tnc from different applications and ssids. This needs to be well tested and we don't want to break the "normal" tnc behavior. Here I need your help with testing, then.


@drowe67 as soon as this PR has been merged, the cli tools for remote control via SSH are ready for usage.

DJ2LS:
> Todo:
> 
> `static.RX_BUFFER_SIZE = config['TNC']['rxbuffersize']` needs to be converted to int ( `parseint ()` ). Maybe we can solve the buffer problem

`TEST_QUEUE = queue.Queue(maxsize="10")` reproduces error. Fix by forcing int type.

DJ2LS:
We need to merge this now, so we have protocols breaking changes published to the world. Adjustements to SSID can be part of smaller releases. This PR became bigger than expected....

PR #327, opened 2023-01-15
Waterfall theme

Mashintime:
Waterfall theme can be changed, tweaks to waterfall for optimization, other small optimizations here and there.

Tested on win and linux (rpi).  

On the rpi CPU utilization is 5-6% higher, but the GUI feels more responsive.

### DIFF
--- a/gui/main.js
+++ b/gui/main.js
@@ -317,7 +317,7 @@ app.whenReady().then(() => {
     }, 3000);
 
     // start daemon by checking os
-    mainLog.info('Starting freedata-daemon binary');
+    mainLog.info('Starting freedata-daemon binary; if it exists....');
 
     if(os.platform()=='darwin'){
         daemonProcess = spawn(path.join(process.resourcesPath, 'tnc', 'freedata-daemon'), [],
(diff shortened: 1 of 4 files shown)


DJ2LS:
@Mashintime 🎉 I will have a look this evening. 

Because of the amount of changes, I suggest we're commuting this to main and not into ls-arq so we can version split this and release it as a separate version. 

I also figured out, the progress bar causes high cpu load. Maybe we can include optimisation at this area into this changes?

DJ2LS:
@Mashintime theres also an "onchange" Part missing for instantly changing the colour theme when selecting a different one. Can you include this? Let me know if you need assistance 🙂👍

Mashintime:
@DJ2LS  Sure, I hadn't gotten that far with profiling.  Are you speaking of the progress bar in the lower right hand corner that renders while in a session?

I'll research rebasing branches; will be a first for me. Should you typically always branch off main?

Regarding the 'on change' for the waterfall, likely related to my inexperience with js, is listening for the click event of the dropdown not enough?

DJ2LS:
@Mashintime 

Yes, no problem, you're doing great!

There's no "always" I think. This depends mostly on circumstances. If you like to contribute to ls-arq for example, you could commit directly or via PR. And if you're working on a new feature for example, it's making sense committing via PR to main I think. At least it's a question of how we want to organise this. 

Regarding to onchange - yes, you're right! There are already other onchange events - just have a look for them and do a copy and paste and adjust them to your needs.
We might consider moving then to an own function, so code is easier to read.

DJ2LS:
Regarding to progress bars: yes, transmission progress and Audio level. We need to reduce cpu load somehow

DJ2LS:
> Regarding to progress bars: yes, transmission progress and Audio level. We need to reduce cpu load somehow 

CQ, Ping. 

But no hurry! 

Also a needed feature: switching between byte/min and bit/s

PR #334, opened 2023-01-18
GUI optimisation #2

DJ2LS:
## Main
- [x] Display tnc ip in title 
- [ ] Reduce progressbar CPU load -> move to chartjs ? ( chart js not needed if we can reduce cpu for bootstrap )
- [x] Use electron progress bar 
- [x] Show version mismatch correctly 
- [x] Move progressbar of transmission to tnc-message instead of state message 
- [x] Notification in settings to restart the tnc 
- [x] Remove Scrollbar from splash screen 
- [x] Disable ping and CQ button when pressed, then show a spinner
- [x] update to latest version
- [x] Eliminate error messages when starting GUI in a dev environment when missing daemon binary
- [x] Fix for heard stations list disappearing when entering text into dxcall textbox
- [x] Clicking an entry in the heard stations list will populate the dxcall textbox with the dxcall
- [x] Update channel not saved anymore - maybe we should also drop a log message on change

## Chat 
- [x] Fix Chartjs error

## Optional challenge 
- [ ] Adjust timeout before transmission


Mashintime (commented):
@DJ2LS Does this help with progress bar CPU usage?

Mashintime (commented):
Tested and works (on windows) if problems exist on other platforms probably because of this.  Was getting tired of seeing all the red error messages in a dev environment.

DJ2LS:
@Mashintime Yes, this should work as a first, partial fix. Next step could be, using inc-messages instead of tnc_state. But this might also need some adjustments to the tnc messages

Mashintime:
Yeah, working on moving it to tnc-messages atm.  Almost there.

DJ2LS:
@Mashintime Update-Channel isn't saved. Can you have a look at this?

DJ2LS:
I hope my possible fix for the update channel isn't causing a crash for you. Sorry was too fast and tired this morning

g0hww:
Would it be possible to add a power spectrum plot, with peak hold? I imagine offering it as a selectable alternative to the FFT waterfall. This would be useful to me, as I have a constant carrier inside the wide mode passband on 40m, and I would be able to assess its strength more easily.

Mashintime:
> 

No problems on my end.

gui/preload-main.js
```diff
@@ -982,8 +983,10 @@ document.getElementById('hamlib_rigctld_stop').addEventListener('click', () => {
 
     // startPing button clicked
     document.getElementById("sendPing").addEventListener("click", () => {
-        var dxcallsign = document.getElementById("dxCall").value;
-        dxcallsign = dxcallsign.toUpperCase();
+        var dxcallsign = document.getElementById("dxCall").value.toUpperCase();
+        if (dxcallsign == "" || dxcallsign == null || dxcallsign == undefined)
+            return;
+        pauseButton(document.getElementById("sendPing"),500);
```
DJ2LS, line 989: I would increasing this to 1000 or maybe 2000.

gui/preload-main.js
```diff
@@ -1018,6 +1021,7 @@ document.getElementById('hamlib_rigctld_stop').addEventListener('click', () => {
 
     // sendCQ button clicked
     document.getElementById("sendCQ").addEventListener("click", () => {
+        pauseButton(document.getElementById("sendCQ"),500);
```
DJ2LS, line 1024: Same here

gui/preload-main.js
```diff
@@ -1944,6 +1959,9 @@ var speedChartOptions = {
         var dxCall = document.createElement("td");
         var dxCallText = document.createElement('span');
         dxCallText.innerText = arg.stations[i]['dxcallsign'];
+        row.addEventListener("click", function(e) {
+            document.getElementById("dxCall").value = dxCallText.innerText;
```
DJ2LS, line 1963: Well done! We should also check if the event listener already exists. In case of updating the row, ir will be created twice or more. 
See preload-chat for example 🙂

gui/package.json
```diff
@@ -8,8 +8,8 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "engines": {
-    "node": ">=14.0.0",
-    "npm": ">=6.0.0"
+    "node": ">=16.0.0",
```
DJ2LS, line 11: Not sure about this. Maybe we can't use it anymore on raspberry / Debian

DJ2LS (commented):
Partially reviewed

gui/package.json
Mashintime: Not sure, it was merged from main from recent updates.

gui/preload-main.js
Mashintime: Is that necessary?  The table gets cleared than recreated on each tnc-state message (as far as I can tell).

gui/preload-main.js
Mashintime: I'll change to 2000

gui/preload-main.js
Mashintime: I'll change to 2000

Mashintime:
> Would it be possible to add a power spectrum plot, with peak hold? I imagine offering it as a selectable alternative to the FFT waterfall. This would be useful to me, as I have a constant carrier inside the wide mode passband on 40m, and I would be able to assess its strength more easily.

@g0hww I just recently ripped out the spectrum plot in hopes of improving performance.  I'd like to get this PR wrapped up, but will look at adding it back with next GUI iteration.

gui/preload-main.js
DJ2LS: Not sure, I only have a small screen (mobile phone) . If it's working, we can keep if, but should keep it in mind in case if problems. Nevertheless a nice feature I was missing as well.

gui/package.json
DJ2LS: Let's keep it as it is and in case of problems we can have another look. I prefer up to date software

gui/preload-main.js
Mashintime: Did some quick research and it sounds the engine with GC the events if it's owner is destroyed.  I think it'll be ok.

tnc/daemon.py
```diff
@@ -369,7 +373,7 @@ def worker(self):
                         rigctld_ip=rigctld_ip,
                         rigctld_port=rigctld_port,
                     )
-
+                    hamlib.set_tx_delay(static.TX_DELAY)
```
DJ2LS: Wrong place. daemon just forwards data to main.py. This might cause the failing ctests

tnc/modem.py
```diff
@@ -236,7 +236,8 @@ class Object:
             rigctld_ip=static.HAMLIB_RIGCTLD_IP,
             rigctld_port=static.HAMLIB_RIGCTLD_PORT,
         )
-
+        self.hamlib.set_tx_delay(static.TX_DELAY)
```
DJ2LS, line 239: this needs to be deleted

DJ2LS:
Suggestion: let's take compression factor from tnc_state first. 

Then making these changes here ready and publish them.

In another PR we can do further adjustments

Mashintime:
> Suggestion: let's take compression factor from tnc_state first.
> 
> Then making these changes here ready and publish them.
> 
> In another PR we can do further adjustments

Ok, I'll revert and revisit.  Also the autoupdate channel appears to be working with the 'on change' change.

gui/preload-main.js
```diff
@@ -1518,7 +1518,10 @@ ipcRenderer.on('action-update-transmission-status', (event, arg) => {
     document.getElementById("bytes_per_min").textContent = arq_bytes_per_minute;
     
     // SET BYTES PER MINUTE COMPRESSED
-    var arq_bytes_per_minute_compressed = Math.round(arq_bytes_per_minute * data.compression_factor);
+    var compress = data.compression_factor;
+    if (isNaN(compress))
```
DJ2LS, line 1522: Nice, I'm wondering if we can use this function for more sanity checks

Mashintime: I believe we can.

DJ2LS:
https://stackoverflow.com/a/9622873

We could add this as own class or maybe this 

https://stackoverflow.com/a/9630132


Also adding the force-gpu css class to waterfall, scatter and chart canvas

DJ2LS:
> https://stackoverflow.com/a/9622873
> 
> We could add this as own class or maybe this
> 
> https://stackoverflow.com/a/9630132
> 
> Also adding the force-gpu css class to waterfall, scatter and chart canvas

I implemented this. Suggestion: creating an option for enabling/disabling fancy effects

Mashintime:
> > https://stackoverflow.com/a/9622873
> > We could add this as own class or maybe this
> > https://stackoverflow.com/a/9630132
> > Also adding the force-gpu css class to waterfall, scatter and chart canvas
> 
> I implemented this. Suggestion: creating an option for enabling/disabling fancy effects

Great, performance on the PI is much improved!  Also thanks for merging after merging ls-arq into main!  I'll see if there's any benefit in disabling the animations.

DJ2LS:
@Mashintime thanks for adjusting the one liners! Readability is important I think!

PR #335, opened 2023-01-19
Hide scroll bar on splash screen

Mashintime:


DJ2LS:
I added you to list of contributors. You should be able now, pushing directly to ls-gui for example

Issue #338, opened 2023-01-22
Emoji/smiley/emoticon chooser in chat window can get stuck and will not close


g0hww:
**Describe the bug**
It is possible to get the smiley chooser stuck open in the chat dialog. Interaction with the chat dialog becomes awkward. It is possible, but not easy, to select another station from the list on the left. The main window is unaffected. CPU load is unaffected. Station seems to function normally.

**To Reproduce**
1. I did insert a smiley into a message I had written, but it didn't go where i wanted. I had set the character insertion point a few words back in the message. 
2. i tried to get out of the situation but couldn't make the smiley chooser close. 
3. now i have clicked on a different station and added smileys into a message i didn't want to write, to a station I didn't want to send a message to. 
4. closing and reopening the chat window doesn't help. 
5. The smiley chooser seems to be permanently stuck on the screen

**Desktop (please complete the following information):**
 - OS: Pop!_OS Linux
 - Version 22.04 LTS
 - Platform x86_64

**Additional context**
Perhaps hitting Esc could make the smiley chooser close, if there are no better ideas. I'm not a GUI expert so I don't know what is normal. Perhaps mouse clicking outside the chooser, too. Maybe closing the chat dialog.

I actually got out of the situation, by sending the message i didn't want to write, to the station I didn't want to send a message to, LOL.

![freedata-bug-smiley-chooser](https://user-images.githubusercontent.com/1413592/213931335-9c880947-36f2-4548-8095-4f443fdb6b3e.jpeg)

DJ2LS:
@Mashintime the chat needs some big improvements. As soon as we have reduced cpu load for the main gui, we need to adjust the chat

Mashintime:
Not able to reproduce, but if it happens again; does clicking the smiley icon (the one that opens the menu) also close the menu?

DJ2LS:
Yes, it does. But we need to add further close actions like we know from modern instant messengers. 
Maybe we should do some brainstorming,first

DJ2LS:
@g0hww can you test this? Just git pull from main.

When clicking on an emoji, the emoji window should close now.

g0hww:
I can now dismiss the smiley chooser by selecting a smiley. That seems to work.
However, once I have done that, and hit enter to transmit the message, it is not sent. The pending message is handled by the UI as if it was sent and moves up into the chat history.
I can then force it to be sent by clicking the little circular arrow to the left of the message in the chat history.
[...]

PR #347, opened 2023-01-29
collision detection experiments and further ARQ improvements

DJ2LS:
Working here on

- collision detection use-case
- fixing bugs
- usability improvements
- code cleanups

PR #349, opened 2023-01-30
Chat window tweaks

Mashintime:
Addresses issue #338, force-gpu and low graphics mode support to chat window progress bars.


Mashintime:
I think the failure checks are related.  But this is ready to merge if approved after review.

DJ2LS:
@Mashintime the last failing build step can be counted as pass. It's because you're on a older build action which has been fixed by #347

DJ2LS (approved):
The emoji picker will be closed on emoji selection? That's okay I think, we can adjust this later if needed by users.

PR #356, opened 2023-02-10
qm-gui-about

Mashintime:
- Added about modal
- Header redesign
- Heard station table tweaks
- Stats can be enabled independent of explorer reporting
- Chat messages have a lower UUID lessening protocol overhead


gui/sock.js
```diff
@@ -542,8 +542,8 @@ exports.sendMessage = function(dxcallsign, mode, frames, data, checksum, uuid, c
     //data = utf8.encode(data)
 
     //socketLog.info(data)
-
-    
+    //shorten uuid
```
DJ2LS, line 545: This comments can be removed?

DJ2LS (approved):
Looks good so far from code point of view need to do a short test to check the gui

DJ2LS:
Gui looks good, thanks @Mashintime 

Suggestions: 
The dropdown field for settings, explorer and so on - I'm not sure if it's too close to start/stop the tnc. Also about the colors. Maybe we should use a grey style für the "chat" as well. Maybe "grey + outline" and moving the Dropdown arrow to the right and making it slightly bigger. Maybe with padding class ( p-2 ) or so. What do you think?
For the dropdown elements I suggest moving the icons to the left and using icons for all elements.

DJ2LS:
note: failing ctest for python 3.12-dev can be ignored

DJ2LS:
image

@Mashintime , what about this?

Mashintime:
Looks good to me.  Commit if you'd like, otherwise can implement later today @DJ2LS

DJ2LS:
@Mashintime I would appreciate, if you could implement it if you like. Just take the code from comment above.

Mashintime:
@DJ2LS no problem!

gui/sock.js
Mashintime: They've been cleaned up.

DJ2LS:
@Mashintime I'm ready for merging this one. So we can test from main together with the other improvements on ls-arq

Mashintime:
@DJ2LS looks good to me

Issue #325, opened 2023-01-02
wrong byte order bug


DJ2LS:
Wrong byte order occurs.

- Burst "A" is getting lost - tnc runs into timeout
- TNC reduced speed level, receiving 2 bursts at lower speed level
- TNC increases speed level, datac1 enabled, old burst "A" will be decoded

Does this occur also with datac3?

```
2023-01-02 11:14:45 [info     ] [TNC] ARQ | RX | SENDING ACK   bytesperminute=1553 finished=3910
2023-01-02 11:14:45 [debug    ] [MDM] [demod_audio] Pushing received data to received_queue nbytes=512
2023-01-02 11:14:45 [debug    ] [MDM] worker_received: received data!
2023-01-02 11:14:46 [info     ] [MDM] calculate_snr:           snr=8.0
2023-01-02 11:14:48 [debug    ] [TNC] enqueue_frame_for_tx     c2_mode=sig0
2023-01-02 11:14:48 [debug    ] [MDM] TRANSMIT                 mode=14 payload=14
2023-01-02 11:14:49 [debug    ] [MDM] ON AIR TIME              time=0.6921732425689697
2023-01-02 11:14:49 [debug    ] [MDM] self.modem_transmit_queue qsize=0
2023-01-02 11:14:49 [debug    ] [TNC] send_data_to_socket_queue: jsondata={"freedata":"tnc-message","arq":"transmission","status":"receiving","uuid":"367e7a21-1169-4180-ae2e-2485cce697d7","percent":1,"bytesperminute":1473,"mycallsign":"EI7IG-0","dxcallsign":"EI3HIB-0","finished":4122}
2023-01-02 11:14:49 [debug    ] [TNC] process_data:            n_retries_per_burst=0
2023-01-02 11:14:49 [debug    ] [HLP] check_callsign: Checking: callsign=b'EI7IG-0'
2023-01-02 11:14:49 [debug    ] [HLP] check_callsign: Checking: callsign=b'EI7IG-0'
2023-01-02 11:14:49 [debug    ] [HLP] check_sessionid: Checking: check=b'\x01' ownid=b'L'
2023-01-02 11:14:49 [debug    ] [HLP] check_sessionid: Checking: check=b'L' ownid=b'L'
2023-01-02 11:14:49 [debug    ] [TNC] RX SNR                   snr=8.0
2023-01-02 11:14:49 [debug    ] [TNC] static.RX_BURST_BUFFER   buffer=[b'aa']
[...]

PR #360, opened 2023-02-13
some ARQ fixes and preparing for TCI

DJ2LS:
- introduce IRS buffer in NACK burst for comparing it with ISS buffer position
- if two stations connect at same time, as ISS we need to ignore the other ISS
- Tests with TCI
- Tests with response request 
- Tests with failing frame crc check


DJ2LS:
@Mashintime i was surprisingly able fixing a minor arq protocol related bug.#325

tnc/rigctld.py
```diff
@@ -182,7 +182,7 @@ def send_data_command(self, command, expect_answer) -> bytes:
                 while recv:
                     try:
 
-                        data = self.data_connection.recv(64)
+                        data = self.data_connection.recv(4800)
```
DJ2LS, line 185: Not sure if this causes problems

DJ2LS:
@Mashintime for some reason this is breaking png image preview
https://github.com/DJ2LS/FreeDATA/blob/ls-arq/gui/preload-chat.js#L446

tnc/rigctld.py
Mashintime: Seems to be working fine on my end.

Mashintime:
> @Mashintime i was surprisingly able fixing a minor arq protocol related bug.#325 - at least I hope so

Nice job@DJ2LS !

Mashintime:
> @Mashintime for some reason this is breaking png image preview https://github.com/DJ2LS/FreeDATA/blob/ls-arq/gui/preload-chat.js#L446

@DJ2LS should be resolved.

dk5sm (approved):
pretty much work. hope, this takes FD to another level of stability.

PR #375, opened 2023-03-16
GUI adjustments

DJ2LS:
The next version of bootstrap ( v 5.3 ) brings us dark mode features and easier designing of custom themes.
This PR tries to add compatibility with v5.3 and making the GUI ready for these versions. 

Todos for this PR:

- [x] Make GUI as much flexible as possible so we can use all bootstrap features natively without customizing and modification. This means, we have to remove all "custom colors" which have been set by css style rules and all used bootstrap color overriding which are not part of the default templates, like the headers of "bootstrap cards".
- [x] Add a design changer for "light/dark/auto"
- [x] Fix smaller GUI related parts which are html/css only

Not Todos for this PR

- no bug fixing of gui related bugs in JS/NodeJS code

image

### DIFF
--- a/gui/package.json
+++ b/gui/package.json
@@ -32,7 +32,7 @@
     "@electron/osx-sign": "^1.0.4",
     "@popperjs/core": "^2.11.6",
     "blob-util": "^2.0.2",
-    "bootstrap": "^5.2.3",
+    "bootstrap": "^5.3.0-alpha.1",
     "bootstrap-icons": "^1.10.3",
     "bootswatch": "^5.2.3",
     "browser-image-compression": "^2.0.0",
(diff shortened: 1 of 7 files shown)


DJ2LS:
@Mashintime @dk5sm 

Bootstrap 5.3 is still alpha, but I expect the version is running stable for our needs. Would you merge this already or waiting some more time until we are going to publish this?

Mashintime:
I almost feel dark mode is not ready.  It does nothing for the default color classes when switching modes (i.e. primary, error, warning, info, etc.).  I think the different themes do this much better than bootstrap's dark mode.  For example RF chat with default/dark mode enabled is still bright.  Switch to theme Slate, and this is what I would envision dark mode to look like.  To support dark mode as is; we'd have to make everything gray.

DJ2LS:
We can wait until the upcoming bootstrap version is more stable and see what kind of changes happen to the color modes. We should keep in mind it's still alpha.1

No problem leaving this PR open and waiting some time

Mashintime (approved):
I'm fine with merging this 'as is' since some of the manual color overrides have been fixed and should help with all the various other themes.  Also the auto dark mode is cool.  ;)

Issue #373, opened 2023-03-12
Frequency control display broken (0.8.0-a.1, Windows)


dk5sm:
After starting the TNC, the frequency control section shows like this - either frequency and/or mode not showing up:

![image](https://user-images.githubusercontent.com/58949681/224530109-6fa8c791-0fd2-4bc3-a8a2-8ec223c6ff91.png)
![image](https://user-images.githubusercontent.com/58949681/224530175-6c817db1-02e3-41bf-b09f-e0de8ea8d39d.png)

Setting frequency or mode manually is working fine, but it seems that reading out does not work:
![image](https://user-images.githubusercontent.com/58949681/224530224-7f074183-dd13-405d-b541-ae0bb61ab394.png)

The behavior was observed under Windows only.

Mashintime:
This info comes from the TNC to the GUI.  I'd guess we'd need to figure out whether it's the tnc or hamlib.  Did you observe the issue with previous versions of FD?

Maybe ensure you have the latest hamlib and/or running manually rigctld with -vvvv and other args needed for FD

DJ2LS:
If it's a version related issue we need to update docs.

I'm running Hamlib 4.4 or so and it's working fine on Ubuntu. Question is, if we can reproduce this with latest Hamlib on Ubuntu as well.

dk5sm:
I do not notice this problem on Ubuntu as well. But it´s reproduceable on mine and another Windows-Box. On both machines hamlib is running in version 4.5.4. Same hamlib version used with the previous FD version 0.7.1 / 0.7.2 / 0.7.3 - but without this issue.
We had a similar gui/hamlib issue before. Maybe an unmatched dependency?

Mashintime:
I don't experience this issue with 4.5.4 on Windows.  Are both windows boxes using the same transceiver/interface?

dk5sm:
No. I am using an IC-705 and on the other machine an IC-7300 is attached. Both via USB-cable.

dk5sm:
![image](https://user-images.githubusercontent.com/58949681/224564962-7149b3d6-61e2-40f8-a043-6752c2f08b9a.png)

I am running rigctld as a windows service on both machines, but running it on the commandline makes no difference.  However, hamlib´s console is throwing an error in an endless loop - "rig_powerstat is not on = 0". Maybe this is a hint?

DJ2LS:
[...]

PR #379, opened 2023-03-20
Fix Issue #373

Mashintime:
Timeout value is too low for some rigctl calls, allow a little more time before timing out.   Addresses issue #373


DJ2LS (approved):
Worth giving a try 🙂

PR #386, opened 2023-03-27
Work on TCI support

DJ2LS:
Further work on TCI support:

- [x] Check for TX CHRONO packages
- [x] TX Processing of tx audio queue if PTT == True and TX Chrono
- [ ] Check if rx audio is working


DJ2LS:
@Mashintime I'd like to merge this one so it's part of the tnc code and therefore it's getting tested by users. It needs some improvements, but we need to wait some time for our TCI tester and his improved antenna situation.

Mashintime (approved):

PR #405, opened 2023-04-25
GUI Tweaks

Mashintime:
Some small quick fixes after watching a few new users use the GUI for the first time.  

https://www.youtube.com/live/-qWuELJRHCg?feature=share

### DIFF
--- a/gui/main.js
+++ b/gui/main.js
@@ -80,13 +80,13 @@ const configDefaultSettings =
                   "screen_height" : 430,\
                   "screen_width" : 1050,\
                   "update_channel" : "latest",\
-                  "beacon_interval" : 5,\
+                  "beacon_interval" : 60,\
                   "received_files_folder" : "None",\
                   "tuning_range_fmin" : "-50.0",\
                   "tuning_range_fmax" : "50.0",\
                   "respond_to_cq" : "True",\
                   "rx_buffer_size" : "16", \
-                  "enable_explorer" : "False", \
+                  "enable_explorer" : "True", \
                   "wftheme": 2, \
                   "high_graphics" : "True",\
                   "explorer_stats" : "False", \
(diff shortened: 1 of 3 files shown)


DJ2LS:
Further thoughts: 
- Change Hamlib "connection possible" to something like "running/connected"
- If running/connected: disable start button for rigctld 
- Add "rigctld" to parameter line 
- Disable input of parameter line 
- if Hamlib selected, make sure it's running or connected, otherwise tnc can't start.
- bring start stop Hamlib buttons closer together 
- Re-Enable info if we're receiving data

DJ2LS:
Not sure about "explorer" and enable it by default because of privacy reasons. What about a first start dialog?

Mashintime:
Ok, I'll revert explorer back to false.  Just thought it would be easier to 'find' new users and switch frequencies and try to make QSOs for us poor NA folks.

Rigctl command line textbox is already disabled with previous commits.  Would prefer not to populate rigctld in textbox since it's small to begin with.  Clicking the copy button populates it all to clipboard though.

Will rearrange rigctld area and like the idea of preventing start if rigtcld is not running and will look at implementing as well as other suggestions in this area.

Not clear what you mean by 'Re-Enable info if we're receiving data'.

DJ2LS:
There was a comment in the video, where it was not clear if a station is still getting data. Specially in times of high Packet loss on poor channels.

Mashintime:
I think all of these have been reasonably addressed, except for the last item.  Not really sure how to make it more obvious no more data is coming in.

DJ2LS:
Thanks for your work, @Mashintime , I wasn't able testing it within app, yet. But all of these changes make sense, I think

I'm wondering if we can make the settings button more visible without bloating navbar too much. I had the feeling they ( the testers in video ) didn't find it that fast. Specially the Hamlib settings have been found too late. 

They would have been able using FreeDATA really fast because they have had a running rigctld in an external terminal. Maybe we can notify the user about this option somehow 🤔

DJ2LS:
@Mashintime How about your progress here? Do we need any more adjustments, or can we merge this one?

Issue #408, opened 2023-05-05
Fix heard stations list


DJ2LS:
<img width="1429" alt="image" src="https://user-images.githubusercontent.com/75909252/236394131-7c23af3d-3841-445d-8ec7-a8fb6bf0fa17.png">

Fix heard stations list if too many entries

PR #411, opened 2023-05-08
Fix issue #408

Mashintime:
Show scroll bar when heard stations list exceeds card height.


DJ2LS (approved):
This could work. I will check this during day when receiving more stations.

PR #412, opened 2023-05-09
Message store and repeat

DJ2LS:
Work on Store and Retry

Workflow:
If we received a beacon or a ping, we will try to send failed messages again.

ToDo:
- Check message if they have a valid state - calculate message length vs timestamp and then set state to false

Side effect: Fixes a tnc related bug


dk5sm (commented):
good idea. first tests were promising

Mashintime (approved):
Looking good to me.

PR #414, opened 2023-05-13
Fix slot based busy detection

Mashintime:
TNC will not go into busy state if slot 0 is not busy as variable gets clobbered.

### DIFF
--- a/tnc/modem.py
+++ b/tnc/modem.py
@@ -1009,6 +1009,12 @@ class RF:
             self.log.debug("[MDM] self.modem_transmit_queue", qsize=queuesize)
             data = self.modem_transmit_queue.get()
 
+            # TODO: carefully test this
+            # avoid sending data while we are receiving codec2 signalling data
+            interrupt_time = time.time() + 5
+            while ModemParam.is_codec2_traffic and time.time() < interrupt_time:
+                threading.Event().wait(0.01)
+
             # self.log.debug("[MDM] worker_transmit", mode=data[0])
             self.transmit(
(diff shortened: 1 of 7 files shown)


tnc/modem.py
```diff
@@ -1251,11 +1251,11 @@ def calculate_fft(self) -> None:
                         range_end = range[1]
                         # define the area, we are detecting busy state
                         #dfft = dfft[120:176] if TNC.low_bandwidth_mode else dfft[65:231]
-                        dfft = dfft[range_start:range_end]
+                        slotdfft = dfft[range_start:range_end]
                         # Check for signals higher than average by checking for "100"
                         # If we have a signal, increment our channel_busy delay counter
                         # so we have a smoother state toggle
-                        if np.sum(dfft[dfft > avg + 15]) >= 400 and not TNC.transmitting:
+                        if np.sum(slotdfft[slotdfft > avg + 15]) >= 400 and not TNC.transmitting:
```
Mashintime, line 1258: @DJ2LS I'm not sure if the >= 400 should/needs to be adjusted since we're looking at a smaller sample size.  Thoughts?

DJ2LS: This needs to be carefully tested I think. We can reduce this step by step. What about reducing it to 200 as a first test?

Mashintime: 200 definitely seems better, but may need further refinement.

DJ2LS:
@Mashintime sorry, I've added a broadcast related commit.

PR #444, opened 2023-05-27
show connected station in chat

### LINKED ISSUE #431: show "connect to" in chat
show "connect to" in chat for easier status check of the tnc. Adding this to top nav bar in chat might be a good position.

DJ2LS:
show connected station in chat closes #431

### DIFF
--- a/gui/main.js
+++ b/gui/main.js
@@ -454,6 +454,8 @@ ipcMain.on("request-update-transmission-status", (event, arg) => {
 
 ipcMain.on("request-update-reception-status", (event, arg) => {
   win.webContents.send("action-update-reception-status", arg);
+  chat.webContents.send("action-update-reception-status", arg);
+
 });
 
 ipcMain.on("request-open-tnc-log", () => {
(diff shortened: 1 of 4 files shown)


DJ2LS:
@Mashintime @LA3QMA could you have a look if this is working more or less correctly? I was busy so I have no idea if this is still working. Feel free improving this if you want

LA3QMA:
Seems to work as expected.

I initiated a chat from STATION1 to HOMEBASE.
The GUI for HOMEBASE says "Connected with: STATION1"

Tried to connect STATION2 to HOMEBASE at the same time and it seems to work ok.

When HOMEBASE receive a message the "Connected with" switch to the last received callsign.
When HOMEBASE sends a return message the "Connected with" switch to the correct callsign.

Should probably be a timeout after a while and remove the "Connected with"

DJ2LS:
Good point with the timeout. @Mashintime do you have the time, implementing a timeout?

The timeout can be a separate PR I think. @LA3QMA could you Sonne the favour and creating a feature request / optimisation for this ?

Mashintime:
Should be easy enough, I can continue the work here if that's ok.

DJ2LS:
@Mashintime yes, this is okay I think.

Mashintime:
@DJ2LS I believe this is ready for main, after your review.

PR #448, opened 2023-06-11
additional message highlighting - os notifications, unread messages, improvements

Mashintime:
- Call list will have an '*' prefixed to call sign and background color set if a new message is available
- Show notifications via the operating system on CQ and messages; configuration option available to disable
- RF Chat button background will change to yellow to indicate a new message/activity is available
- Minor console.log clean up in chat module
- Database pruning option added to clean older pings and beacons out of the database
- Ignore unusable flex radio audio devices (there are just too many)

Ready for testing and feedback.


gui/main.js
```diff
@@ -100,7 +100,8 @@ const configDefaultSettings =
                   "max_retry_attempts" : 5, \
                   "enable_auto_retry" : "False", \
                   "tx_delay" : 0, \
-                  "auto_start": 0 \
+                  "auto_start": 0, \
+                  "notification": 1 \
```
DJ2LS, line 104: Can we name this more self explaining like "enable_system_notification" or something like that?

tnc/audio.py
```diff
@@ -76,6 +76,9 @@ def fetch_audio_devices(input_devices, output_devices):
         # Use a try/except block because Windows doesn't have an audio device range
         try:
             name = device["name"]
+            # Ignore some Flex Radio devices to make device selection simpler
+            if name.startswith("DAX RESERVED") or name.startswith("DAX IQ"):
```
DJ2LS, line 80: Nice improvement 🙂 Thanks!

DJ2LS (commented):
Nice improvements! Question: Can we count unread messages and displaying this somehow? Also instead of just using "*" can we use a nice gui element or badge as well?

gui/main.js
Mashintime: Updated to enable_sys_notification.

tnc/audio.py
Mashintime: Quality of life improvements are always a plus.

Mashintime:
> Nice improvements! Question: Can we count unread messages and displaying this somehow? Also instead of just using "*" can we use a nice gui element or badge as well?

We could count unread messages, but personally feel it's over kill.  I feel just knowing there are new messages is enough.  Plus no real further GUI elements to add or jumble up the interface is a plus in my opinion.

The '*' was just a lazy/quick implementation.  The callsign area is already so busy, but if you have an idea let me know.

DJ2LS:
We could use it as it is for now and if we have some users requesting this we could think again about it.

Mashintime:
The * should clear on first click.  If you get an error in console let me know.  The chat button doesn't update immediately and is delayed to limit cpu load.

DJ2LS:
@Mashintime ah its working! I didnt see there was another unread message....

DJ2LS:
@Mashintime After some days of testing I would suggest, that a beacon or a ping isnt triggering "unread" message flag. Only messages should mark a chat as "unread".  I'm also wondering about sort oder of unread messages and maybe we can make the background a different color for additional highlighting

Mashintime:
I'll remove the new flag from ping and beacon, easy enough.  Like the idea for changing background color.  Should we do background and * or do you think background would be enough.

I'll look at sort order again, but really won't matter unless user refreshes.

DJ2LS:
I think both is fine. Then we have a text indicator and a colour indicator as well.

Mashintime:
@DJ2LS feel this is ready for merging after your review.

DJ2LS:
@Mashintime looks good so far, question:
is the delete database button just for the selected chat or for the entire database? If so, I suggest, we should add it to app settings instead to chat dropdown.

DJ2LS:
And I think we could add pings to cleanups as well

DJ2LS:
Oh, and @Mashintime , can you update bootstrap from alpha to 5.3.0 ? We have a stable release now

Mashintime:
> Oh, and @Mashintime , can you update bootstrap from alpha to 5.3.0 ? We have a stable release now

No problem

> And I think we could add pings to cleanups as well

Will do

> is the delete database button just for the selected chat or for the entire database? If so, I suggest, we should add it to app settings instead to chat dropdown.

It's for the entire database.  Will move to settings.

DJ2LS:
> > Oh, and @Mashintime , can you update bootstrap from alpha to 5.3.0 ? We have a stable release now
> 
> No problem
> 
> > And I think we could add pings to cleanups as well
> 
> Will do
> 
> > is the delete database button just for the selected chat or for the entire database? If so, I suggest, we should add it to app settings instead to chat dropdown.
> 
> It's for the entire database. Will move to settings.

I think then we can merge this

DJ2LS (approved):
Looks good so far, great work!

PR #449, opened 2023-07-02
WIP: mesh networking prototype

DJ2LS:
This is a bleeding edge prototype for doing mesh networking experiments. 

Actual WIP
- [ ] Routing Table
- [ ] Signalling Table

Signalling and routing table are most important in this area, so we will make sure, they are working before trying to send data.


DJ2LS:
@Mashintime this is the PR you can contribute to, if you want to work on "enable / disable" feature we talked about.

DJ2LS:
Nice, thanks for working on enable / disable mesh feature. We also need an option for starting the tnc with or without a mesh option enabled.

PR #458, opened 2023-07-09
adjusting mesh signalling ping

DJ2LS:
adjusting mesh signalling ping and improving it.

known problems:

- [x] receiving a ping request isn't triggering a ping-ack
- [x] origin isn't known
- [x] router flag is useless --> move this to origin


DJ2LS:
@Mashintime I suggest let's continue work here with mesh related gui changes as well.

Issue #466, opened 2023-08-08
broadcast message length are not reset 


DJ2LS:
when selecting a braodcast and then a message, the input field length is not reset correctrly

PR #472, opened 2023-08-27
Various bug fixes

Mashintime:
- Fix for messages stuck in transmitting state not resetting to failed after a few hours
- Fix for message input being limited when clicking on broadcast messages then a different station #466 
- Fix for waiting messages being tx'd when refreshing view with beacons enabled
- Rework channel busy logic
- Set channel busy if codec2 data is detected but below the detection threshold

### DIFF
--- a/gui/preload-chat.js
+++ b/gui/preload-chat.js
@@ -1060,7 +1060,7 @@ update_chat = function (obj) {
   }
 
   // check if wrong status message
-  if (obj.status == "transmit" && obj.type == "transmit" && obj.percent < 100) {
+  if (obj.status == "transmitting" && obj.type == "transmit" && obj.percent < 100) {
     var TimeDifference = new Date().getTime() / 1000 - obj.timestamp;
     if (TimeDifference > 21600) {
       //Six hours
--- a/gui/preload-chat.js
+++ b/gui/preload-chat.js
@@ -827,6 +827,10 @@ ipcRenderer.on("action-new-msg-received", (event, arg) => {
 
       add_obj_to_database(obj);
       update_chat_obj_by_uuid(obj.uuid);
+      // check for messages which failed and try to transmit them
+      if (config.enable_auto_retry.toUpperCase() == "TRUE") {
+        checkForWaitingMessages(obj.dxcallsign);
+      }
 
       // handle ping-ack
     } else if (item.ping == "acknowledge") {
@@ -865,7 +869,10 @@ ipcRenderer.on("action-new-msg-received", (event, arg) => {
       obj.new = 0;
       add_obj_to_database(obj);
       update_chat_obj_by_uuid(obj.uuid);
-
+      // check for messages which failed and try to transmit them
+      if (config.enable_auto_retry.toUpperCase() == "TRUE") {
+        checkForWaitingMessages(obj.dxcallsign);
+      }
       // handle ARQ transmission
     } else if (item.arq == "transmission" && item.status == "received") {
       //var encoded_data = atob(item.data);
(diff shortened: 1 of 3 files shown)


gui/preload-chat.js
```diff
@@ -1060,7 +1067,7 @@ update_chat = function (obj) {
   }
 
   // check if wrong status message
-  if (obj.status == "transmit" && obj.type == "transmit" && obj.percent < 100) {
+  if (obj.status == "transmitting" && obj.type == "transmit" && obj.percent < 100) {
```
DJ2LS, line 1070: I also had this for messages, which are at 100%. But we can test it with this approach at first.

tnc/modem.py
```diff
@@ -875,7 +875,10 @@ def demodulate_audio(
                         # we need to disable this if in testmode as its causing problems with FIFO it seems
                         if not TESTMODE:
                             ModemParam.is_codec2_traffic = True
-
+                            if not ModemParam.channel_busy:
+                                self.log.debug("[MDM] Setting channel_busy since codec2 data detected")
+                                ModemParam.channel_busy=True
+                                ModemParam.channel_busy_delay+=10
```
DJ2LS, line 881: Good idea, and important for improving the channel timeouts ( another PR )

tnc/modem.py
```diff
@@ -1261,7 +1264,7 @@ def calculate_fft(self) -> None:
         whether the channel is "busy."
         """
         # Initialize channel_busy_delay counter
-        channel_busy_delay = 0
+        #channel_busy_delay = 0
```
DJ2LS, line 1267: I'm not sure if this is causing a possible crash?

tnc/modem.py
```diff
@@ -1354,22 +1360,23 @@ def calculate_fft(self) -> None:
...
-                            # When our channel busy counter reaches 0, toggle state to False
-                            if channel_busy_delay == 0:
-                                ModemParam.channel_busy = False
-                                ModemParam.channel_busy_slot[slot] = False
-
+                            ModemParam.channel_busy_slot[slot] = False
                         # increment slot
                         slot += 1
-
+                    if (addDelay):
```
DJ2LS, line 1369: This will possibly not be reset, because we don't have a "addDelay=False" . Question: Why is this in a separate if clause and not part of "if np.sum(slotdfft[slotdfft > avg + 15]) >= 200 and not" ? 

Is there are reason for putting "channel_busy_delay" as a global ( ModemParam.channel_busy_delay )?

DJ2LS (changes requested):
Looks good so far, but the TNC code needs some fixes I think.

gui/preload-chat.js
Mashintime: Resetting messages stuck in transmitting and equal to 100 should be reset after 6 hours when refreshing or relaunching already.  Should be close to this code block.

tnc/modem.py
Mashintime: No  ModemParam.channel_busy_delay is storing this value with this PR.  We need to be able to set delay in the codec2 detection (see above comment code section)

tnc/modem.py
Mashintime: addDelay gets reset every time the while loop restarts.

It's outside of the for loop because during slot detection, it could get triggered for each slot on a single pass.  You could have a delay of 50 if each slot is busy on a single pass, but also the delay will be lessened quicker if each slot is not busy.

Mashintime: The intent is to only increase delay if one or more slots are busy, or decrease delay if all slots are not busy.

DJ2LS: Ah now I got it, sorry, didnt see the entire code because of github

DJ2LS (approved):

PR #501, opened 2023-10-24
work on universal builds for macOS

DJ2LS:
https://www.electron.build/configuration/mac


DJ2LS:
⨯ Detected file "Contents/Resources/modem/_internal/Python.framework/Versions/3.11/Python" that's the same in both x64 and arm64 builds and not covered by the x64ArchFiles rule: "${os}/*"  failedTask=build stackTrace=Error: Detected file "Contents/Resources/modem/_internal/Python.framework/Versions/3.11/Python" that's the same in both x64 and arm64 builds and not covered by the x64ArchFiles rule: "${os}/*"

PR #504, opened 2023-10-25
some GUI Adjustments

DJ2LS:
- [x] Fix of json bug in Modem
- [x] Reset message state to failed on startup or after a specific time if message is N minutes old
- [x] Fix audio tune


DJ2LS:
@Mashintime could you have a look at the contributors list?

PR #510, opened 2023-10-28
arq improvements

DJ2LS:
We need to adjust the ARQ protocol and making it more efficient. This PR is a first start.

### Todos:

- [ ] Move from busy-wait to event based system where possible ( while Modem.transmitting)
- [x] Adjust opening data channel
- [ ] Adjust opening of arq session
- [ ] Adjust ACK/NACK of ARQ transmission
- [x] Add busy cool down time to busy slots
- [x] Interrupt busy state if speed level fits to busy condition
- [x] implement morse indicator if needed


DJ2LS:
Thought: 

We can improve efficiency under bad channel conditions, if we are sending a NACK not just after the timeout was running out. We could check about 2-3 seconds, if we are receiving a codec2 signal from ISS. If yes, then ISS received the NACK and is resending data. If not, NACK is probably lost. 
If we are not receiving a codec2 signal after 2-3 seconds we could early interrupt the burst timeout and sending a NACK earlier. 

There is a chance we have a false positive. But if we didn't receive a signal over the full burst transmission time, then we can expect a frame to be lost. 

I think we should try this.

PR #518, opened 2023-11-01
Move radio port setting to hamlib settings

ptsmonteiro:


DJ2LS:
Question is, how we can handle this one later as usage is affected. But we can park it for now in settings

DJ2LS (approved):

PR #520, opened 2023-11-01
Move audio device settings to modem settings area.

ptsmonteiro:
This moves the audio device settings from the main screen to the modem settings tab.


DJ2LS (approved):
I would count this as a thing we should keep in mind and only park it in settings. Maybe we find a way how to make this easier to configure/validate on startup

PR #529, opened 2023-11-04
Some fixes, tweaks

Mashintime:
- Fix newchat (not sure if it's the best method)
- Sort heard stations by timestamp
- Theme support for waterfall


Mashintime:
This can be merged into develop if approved

DJ2LS (approved):
From my side it's okay, but didn't test its.

@Mashintime @ptsmonteiro i'd like to merge this so we have all work merged to develop. Then we can start working on gui design in my opinion.

PR #638, opened 2024-02-11
fix loading default config

arodland:
without this change, if the user doesn't have a valid config on disk, they just get a white screen on startup due to `nconf.required` throwing an exception.


DJ2LS:
@arodland thanks for the PR! We appreciate all support we get!

DJ2LS:
@arodland I'm not sure if this really fixes the problem, as the default config is stored in a variable :thinking: I'll try testing a bit

DJ2LS:
youre right, this fixes the problem!

PR #647, opened 2024-02-11
Internal rigctld tweaks for windows users

Mashintime:
- Process custom arguments for rigctld for all platforms
- Search c:\program files*\Hamlib*\bin for rigctld binaries on Windows
- Remove default custom args from example config

### DIFF
--- a/modem/config.ini.example
+++ b/modem/config.ini.example
@@ -21,7 +21,7 @@ ip = 127.0.0.1
 port = 4532
 path = 
 command = 
-arguments = --cenas
+arguments = 
 
 [RADIO]
 control = disabled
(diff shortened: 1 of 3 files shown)


DJ2LS:
Great, thanks @Mashintime  
What about removing the 32bit version of Hamlib from build process ( nsis )?

@dk5sm might this fix your path problem you've reported?

Mashintime:
NSIS?  We should only support 64 bit in my opinion for binary packages.  

But if a user wants to use a 32 bit version of hamlib, I see no issues with supporting that and checking if it's available.

DJ2LS (approved):

DJ2LS:
@Mashintime I've removed the 32bit build separately. We can merge if you want

Mashintime:
I'm good with merging.

PR #657, opened 2024-02-20
fixed radio control

DJ2LS:
Radio control needed some adjustments

Addresses https://trello.com/c/rFkwbhAB/77-changing-frequencies-also-changes-mode-edit-added-possible-fix


dk5sm:
Pulled latest code from _develop_ and tested with IC-7300 @40W out: when gui starts, frequency and mode is shown correctly, but "% power" is always set to 10. Changing power setting via gui seems to work properly, as trx follows this setting.

PR #685, opened 2024-03-13
Gui fixes

Mashintime:
- Fix 17m frequency in frequency list
- Clicking a station in the heard list(s) will update the ping textbox(s) with origin callsign


Mashintime:
Can be merged into develope if approved.

DJ2LS (approved):
I'm not sure about the event listeners and if there is a Vue related best practice. But if we are both fine with this we can do its way, making the gui more intuitive. We also need to adjust the gui from time to time ..

PR #719, opened 2024-05-05
Handle None responses from rigctld strength command

Mashintime:
- Receiving constant errors in modem as trying to typecast to int without checking value first.


Mashintime:
Ready for merging.  Tested with and without a value and works as expected.

DJ2LS:
:+1:

PR #825, opened 2024-09-30
Install for Fedora 40

OH1KH:
This install script addition seems to run with Fedora 40

### DIFF
--- a/tools/Linux/install-freedata-linux.sh
+++ b/tools/Linux/install-freedata-linux.sh
@@ -117,6 +117,21 @@ case $osname in
 	esac
    ;;
 
+   "Fedora Linux")
+	case $osversion in
+	   "VERSION_ID=40" )
+		sudo dnf install -y git cmake make automake gcc gcc-c++ kernel-devel wget portaudio-devel python3-pyaudio python3-pip python3-colorama python3-virtualenv google-noto-emoji-fonts
+	   ;;
+
+	   *)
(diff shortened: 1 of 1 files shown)


DJ2LS:
@deejdeej could you do me the favour and reviewing it? You're most experienced with the scripts

deejdeej:
> @deejdeej could you do me the favour and reviewing it? You're most experienced with the scripts

Yes, when I have the time I will review it and test it out.

jstrahle:
Just tested the installation with the modified script, and it looks like python-devel should be added to the list of packages being installed.

OH1KH:
Possible.
I did not test script in plain Fedora40 (that needs setting up new virtual machine).
Just seek corresponding Fedora packages for Debian/Ubuntu names.

As I have done other programming with this F40 there might been already loaded packages and perhaps that's why script passed through without python-devel loading.

With Fedora it is also possible to install 
sudo dnf groupinstall "Development Tools" "Development Libraries"

That is some kind of equivalent of the Debian build-essential package, but there comes also lot of unneeded stuff.

jstrahle:
I will test installation with "plain" Fedora 40, and give input later.

deejdeej:
Added support for Fedora 40 into the install script.  Tested on fresh Fedora 40 Workstation install.  Thanks for the code bits and hints @OH1KH and @jstrahle !

PR #860, opened 2025-01-03
Docker image

g7ufo:
I've added documentation at `README.docker.md` for anyone to help test this. ~~The [documentation](https://github.com/DJ2LS/FreeDATA/commit/c8907518667763f3a7d24b2f5ffec127949693ac) points to my forked image for now but will be retired in favour of this repo after merge.~~

The Image build takes about a hour and so is only performed when a tag (I.e from a release) is pushed.

This image isn't MacOS specific, the image is Linux based for `amd64` and `arm64` platforms. The documentation leans towards MacOS but only for the audio setup and the change of port to `5050` (because on MacOS port `5000` is used by 'Airplay Receiver').


DJ2LS:
Thanks a lot for your help with this, @g7ufo .

Some Suggestions: 
- [ ] can we change the paths to the origin repository? 
- [ ] Could you merge into develop branch? It's our central point for publishing releases 

Some inline comments will follow

g7ufo:
> Thanks a lot for your help with this, @g7ufo .
> 
> Some Suggestions:
> 
> * [ ]  can we change the paths to the origin repository?
> * [ ]  Could you merge into develop branch? It's our central point for publishing releases
> 
> Some inline comments will follow

Both comments addressed in last commit.

DJ2LS:
@g7ufo Did you see the other comments above?

Best regards, Simon

g7ufo:
@DJ2LS I can't see any additional comments, there are none inline.

DJ2LS:
> @DJ2LS I can't see any additional comments, there are none inline.

Maybe the comments are hidden. They should be there. You could also look into changed file here in GitHub. Maybe they are visible there

g7ufo:
@DJ2LS Could you maybe send me a screengrab, would be great to get any comments addressed.

One thing which always catches me out with GitHub [at work] is forgetting to hit the 'Submit review' button under 'Review changes'.

DJ2LS:
No worries, we will handle this :-) 

At least at my question was, the ip addresses inside the configs - are they fixed because of docker or can we make them more dynamic or common? Specially rigctld for example

DJ2LS:
Question: what Happens in case of an Update? What's with the Message data and settings?

g7ufo:
@DJ2LS Sorry for the delay in replying. To answer your questions:

> the ip addresses inside the configs - are they fixed because of docker or can we make them more dynamic or common? Specially rigctld for example

The `192.168.0.10:6701` address can (and should) be updated to reflect the IP of the radio connected alongside the port and appropriate model number. I've pushed a change to the documentation to make this clearer.

> Question: what Happens in case of an Update? What's with the Message data and settings?

The `docker` and `docker compose` commands in the documentation both map the configuration directory in the running container to the host machine. Providing the update doesn't make any breaking changes to the config or other data files then it should be seamless.

DJ2LS:
Thanks @g7ufo , so this means, the config folder is separated and in case of a FreeDATA update, the config files and database remains, right? Sorry, I'm not that familiar with docker. But I also want to merge this commit.

g7ufo:
Yes, @DJ2LS  that's correct - the config folder is a mount from the host into the container. Just to illustrate, this is the `data` directory on my host machine:

```
❯ ls -l data
total 56
-rw-r--r-- 1 neil staff  1107 Jan  3 15:05 config.ini
-rw-r--r-- 1 neil staff 53248 Jan  3 15:05 freedata-messages.db
```

So if I use an updated version of the image, this same directory will be mounted in that.

DJ2LS:
Ah, good to know, thanks for taking the time doing this. 
At least we want to have things as simple as possible. So intention is, just downloading a docker container and running it, without too much work.

DJ2LS:
I like to give it a try, so we are going to merge it into develop branch. Then it's part of the next release. I would appreciate if we can continue work on this so everything is as easy as possible.

DJ2LS:
oh, and I don't want to crash your work - but do we really need pulse audio on macOS? We are using the port audio library, it should support macOS natively

Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+
+ADD https://github.com/drowe67/codec2.git ./codec2
+
+WORKDIR /app/FreeDATA/freedata_server/lib/codec2
+
+RUN mkdir build_linux
+
+WORKDIR /app/FreeDATA/freedata_server/lib/codec2/build_linux
+
+RUN cmake .. && make -j4
```
petrkr, line 56: You can specify make codec2, so it will not compile other binaries, which are not used. it will speed up build and safe space

check Arch linux PKGBUILD

https://aur.archlinux.org/packages/freedata-server

g7ufo: Thanks @petrkr that will help! Updated in latest commit.

g7ufo:
> I like to give it a try, so we are going to merge it into develop branch. Then it's part of the next release. I would appreciate if we can continue work on this so everything is as easy as possible.

Yes, absolutely - I'm happy to maintain this image.

> oh, and I don't want to crash your work - but do we really need pulse audio on macOS? We are using the port audio library, it should support macOS natively

No problem at all. To be honest, the audio part was the hardest for me as I've never done that in a container before. I'm not sure how audio would from from host to container without pulse audio - happy for anyone with more experience to make suggestions!

DJ2LS:
Okay, then let's give this a try with next release. Then we might have to adjust some things

DJ2LS:
@g7ufo @petrkr can I merge for testing this in next release?

PR #865, opened 2025-01-22
WIP: connected mode via socket interface

DJ2LS:
This PR contains a socket interface and a connected mode ( in development ). It is under development and not working, yet. **The release date is unknown**. This PR should help tracking progress and getting a better feeling and understanding of this feature.


freedata_server/p2p_connection.py
```diff
@@ -217,51 +217,50 @@ def connected_irs(self, frame):
         self.is_ISS = False
         self.orign = frame["origin"]
         self.destination = frame["destination_crc"]
-
-        if self.socket_command_handler:
-            self.socket_command_handler.socket_respond_connected(self.origin, self.destination, self.bandwidth)
-
+        #If these 2 lines are not here, the receiving station does not reply back with an ACK to a P2P_CONNECTION_CONNECT packet. Is this intentional? Leaving here for testing for now.
         session_open_frame = self.frame_factory.build_p2p_connection_connect_ack(self.destination, self.origin, self.session_id)
```
DJ2LS, line 221: Yes this is making sense to me. Let's use these inline comment feature, as we can mark discussions as solved for example

DJ2LS:
@modernham i suggest before working on specific features like the callsign stuff, we should maybe ensure, the basic transmission is working. So just sending data to a specific callsign, provided bis CONNECT command. Additional stuff like ignoring not registered callsign could be added afterwards.

DJ2LS:
Regarding ARQ inside the connection- my idea was, sending single small data bursts, acknowledging them via ACK burst. As soon as we have data, which doesn't fit into a single burst, we are calling the existing ARQ protocol also used for message transmission, which is doing then the entire work. 

By doing it like this, I wanted to have a trade off between reliability and efficiency for smaller data, but also having a reliable protocol for bigger data.

DJ2LS:
Hi, I merged changes from develop into this branch - so don't forget doing a git pull before committing for avoiding issues :-)

DJ2LS:
I'm going to merge this into develop so we don't have too many breaking changes - as soon we have continuing work here we can open the PR again.

PR #875, opened 2025-01-31
v0.16.11

DJ2LS:


DJ2LS:
@deejdeej there is a style related error with CodeFactor regarding installation scripts - do you have the time having a short look and probably fixing it?

deejdeej:
> @deejdeej there is a style related error with CodeFactor regarding installation scripts - do you have the time having a short look and probably fixing it?

The syntax of that line looks fine.  I do not know why they say it is "legacy" and needs to be changed.  Can you simply whitelist the two scripts?

PR #879, opened 2025-02-02
macOS install Scripts v0.0.1

HB9HBO:
This are the mentiond install and run script for macOS. They depend on brew or MacPorts. They are tested on macOS 14 and macOS 15. A README.md file with the instructions is included.


DJ2LS:
Thanks! I will test them tomorrow :-)

PR #889, opened 2025-02-19
Move API Endpoints to subfolders

DJ2LS:
We need to reorganize the api endpoints so they are more structured and easier to read.


DJ2LS:
@LA3QMA I think this work should make documentation easier as every endpoint has now its own file

PR #903, opened 2025-03-03
Added Czech translation

petrkr:
I have not so much idea if I have to use "cz_Czech" or translate that "Czech" as I did it..

Since in GUI is visible filename and not language name in some applications in Language selector is Czech, in some "Cestina"

Maybe translated names is better for non-english speaking peoples ? Japanese would be then like `jp_日本語.json` instead of `jp_Japanese.json`

### DIFF
--- a/freedata_gui/src/locales/cz_Čeština.json
+++ b/freedata_gui/src/locales/cz_Čeština.json
@@ -1,0 +1,15 @@
+{
+    "settings": {
+        "enable": "Aktivovat",
+        "default": "Výchozí",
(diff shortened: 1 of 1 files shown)


DJ2LS:
Thanks, @petrkr , I think using translated names might be better, as you already said. I think that's something we can change easily afterwards if it doesn't fit. 

How is using the nested keys for you? Makes that sense? Then we could continue with this way and adding more strings. I didn't add more yet, because I want to clarify such things, first :-)

PR #909, opened 2025-03-11
remove tci support 

### LINKED ISSUE #114: TCI support
We need to finish implementation of TCI so it's at least controllable via GUI.

Todos:
- [x] Add TCI field next to Hamlib 
- [x] Add Startup via Daemon
- [ ] Change Hamlib related functions to "radio" related function and variable names 

<hr>

https://github.com/maksimus1210/TCI/blob/master/doc/TCI_interface_1.6.pdf

I hope that the TCI protocol is implemented relatively soon in the project:

TCI – Transceiver Control Interface is the network interface for control, data transfer and synchronization between a transceiver/receiver, logger, digital software, skimmer, other software and external power amplifier, band filters, antenna switch, radio station controller and other devices.

https://github.com/maksimus1210/TCI

Software like MSHV, JTDX etc is using this. It's not limited to SunSDR tranceivers. 

Using TCI you don't need any external/thirdparty VAC as this goes trough websocket

73 de LA3QMA
Kai Gunter

DJ2LS:
Since TCI is only used by a very few persons, this PR won't affect lots of users. closes #114


DJ2LS:
@LA3QMA FYI - Let me know when we can merge this one.

PR #915, opened 2025-03-17
fixed wavelog and made the popup a little better

LA3QMA:
fixed wavelog and made the popup a little better

### DIFF
--- a/freedata_gui/src/js/eventHandler.js
+++ b/freedata_gui/src/js/eventHandler.js
@@ -156,28 +156,26 @@ export function eventDispatcher(data) {
     case "message-logging":
       if (data.endpoint === "wavelog") {
         if (data.status === true) {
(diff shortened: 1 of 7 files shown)


freedata_gui/src/js/eventHandler.js
```diff
@@ -153,6 +153,36 @@ export function eventDispatcher(data) {
       loadAllData();
       return;
 
+    case "message-logging":
+      if (data.endpoint === "wavelog") {
+        if (data.status === true) {
+            const message = `
+            <div>
+                <strong>${i18next.t('Wavelog posted')}:</strong>
```
DJ2LS, line 161: The strings should be translatable, so we need a reference in translation files.

DJ2LS (approved):

PR #987, opened 2025-06-02
R&D audio streaming

DJ2LS:
some testing with streaming audio.

This is a research and development PR for gathering experience with audio streaming. There is no release date. however, it can probably be improved until production readiness. 

Things to find out:
- Best practise for streaming audio to gui and back
- Best parameters for audio buffering in terms of stability and delay
- Test if codec2/rade mode implementation is making sense 
- Useful feature implementation


DJ2LS:
Audio streaming via tcp is problematic. 
For LTE connections we need big chunks, like 2400bytes for example.

For fast network / wifi we could reduce down to 300bytes

DJ2LS:
https://github.com/fastapi/fastapi/issues/960


Problem is TCP/IP which causes problems on slow connections

PR #989, opened 2025-06-23
R&D NORM Groupchat

### LINKED ISSUE #324: Many to Many chat.
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Multi-party chat. i.e. many to many.

### LINKED ISSUE #418: Erasure coding for short message broadcast
Inspired the approach use within the Nack Oriented Reliable Multicast (NORM) protocol, I would like to propose the adoption of an erasure coding strategy, applied to a sequence of codec2 frames and operating at the frame layer, with erasures applying to entire missing frames..
This will use a shortened Reed-Solomon block code in GF8.
I will attach proof-of-concept python code to show what I mean, in a series of follow up posts.

### Assumptions
Use codec2 DATAC4 with 56 bytes of payload per frame.

Allow FreeDATA TNC 22 bytes to identify this message, based on
- SourceId,
- DestId,
- Message Number (perhaps 16-bit int)
- Client Service Data Unit (CSDU) size (7-bit int), max 128 bytes.

This is the BlockHeader BH.

Assume that the sender commits to storing each block for some duration,
perhaps 24 hours, or until the Message Number wraps.

The Erasure Coding Layer needs to identify, within each Erasure Coded Segment,
- The maximum number of segments for the message,  4 bit unsigned
- This Segment number, SegId, 4 bit unsigned
- Number of parity segments, EcSegs, 4 bit unsigned
- Reserved, 4 bit unsigned.

This is the EC Header EH. It is 2 bytes in size.

So, 56 -22 -2 = 32. The ECDU is divided into five 32 byte segments.

We will assume that, by default, we are going to send  a maximum number of
5 frames of DATAC4.
Of that, up to 4 frames will be client service data segments and 1 frame will contain a proactive parity data segment. We will generate more parity data than this, so we could increase the number of parity blocks we send proactively.
This should take 5*5.17 = 25.850 seconds of air time and should be able to convey client service traffic of up to 128 bytes.
Let us choose to generate 4 blocks of parity data, 4*32=128 bytes.

If we were to send the whole thing in one go, we would have a sequence like

- D1,D2,D3,D4,P5,P6,P7,P8

but we plan only to send ( at most)

- D1,D2,D3,D4,P5

initially, retaining P6,P7,P8 for reactive EC responses. These could be transmitted in response to EC Repair Requests, issued by receivers that want to fix errors in received data. It is preferable to respond with parity segments, as any station receiving an additional parity segment will be able to repair any missing data segment with it.

### LINKED ISSUE #863: [Feature Request]: Broadcast messages
### Problem Description

The broadcast message will be sent to all participants who can receive it. This type of message will not send ACK back to the sender. The list of broadcasted messages can be placed on the dashboard and in the chat window as the first chat group.

### Proposed Solution

Add a capability to sent broadcast message to all participants without ACK.

### Alternatives Considered

NA

### Additional Information

_No response_

DJ2LS:
This PR is for R&D for working on a **NORM** ( NACK Oriented Reliable Multicast https://en.wikipedia.org/wiki/NACK-Oriented_Reliable_Multicast ) Groupchat and for getting a feeling about the needs for database structure and the protocol itself. 

⚠️ **Developer Notice**  
This Pull Request is **under active development**.  
Please do **not use or merge** unless you are directly contributing to this feature.
⚠️ 

Goal is, working constantly on this one, while we keep up2date with latest code in `develop` branch. This PR is considered to be a midterm work package. 

Because of the limited nature of RF transmissions, we can't implement NORM 1:1, but we can let us inspire. 

### How it should work: 

- Station A sends data "Hello World....^snip^". This data will be split into bursts of fixed sizes. In this case we talk about 3 bursts, but up to 15 are possible
- Every burst contains the full message header

- Station X receives burst 1 and 3
- Station Y receives burst 1 and 2
- Station X and Y requesting retransmission of burst 2 ( X ) and 3 ( Y ) via broadcast request
- Station A receives the request of Station X and sends burst 2
- Station X receives the request of station Y and sends burst 3
- Station X and Y both receiving the corresponding data and all data can be assembled internally - success
- Meanwhile Station Z also heard the burst retransmissions and continues with the same requesting procedure .... 

### Todos: 

- [x] Basic frame structure for NORM 🚧 
- [x] Database parts 🚧 
- [x] API Endpoints 🚧 
- [ ] Full test covered 🚧 
- [ ] GUI🚧 
- [x] NORM: Request missing data🚧 
- [x] NORM: Transmit requested data🚧 
- [x] NORM: Define a burst lifetime, until we resend data🚧 
- [ ] NORM: Define message types which make sense for broadcasts, have a log at EMCOM needs as well🚧 
- [x] FIX: Correct message type by flag - we have to get a string from flag-int
- [ ] NORM: Handle priority correctly  with schedule managers
- [ ] API: transmissions are blocking, we have to make them asynchronous


DJ2LS:
Stuff to think about:

How do we handle start of repeating messages? 
- Should we do this cyclic via schedule manager, maybe checking for `assembling` status in database combined with last received timestamp?
- Should we trigger this after a specific time, bursts have been received? Maybe calculating some time after a burst has been received, and the time left until total bursts have been processed, then sending a repair burst?
- What's the time to life for a broadcast until we don't answer repair requests anymore?

DJ2LS:
Options: 
Enable / Disable broadcasts
Send repair requests 
Broadcast Lifetime 
Respond and listen only to specific broadcast domains

PR #1017, opened 2025-10-02
Added package-lock.json, made gui build reproducible

as3ii:
Only node libraries can avoid having a package-lock.json. For applications it is required for better reproducibility.
See the official documentation: https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#description

> It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.
>
> This file is intended to be committed into source repositories, and serves various purposes:
> 
> - Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> - Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
> - Facilitate greater visibility of tree changes through readable source control diffs.
> - Optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.


as3ii:
changed PR's base branch

DJ2LS:
I didn't add package-json.lock to the repository, as it will be generated automatically. Question is also, if we add a *.lock file, do we have to update the dependencies manually then? Or wouldn't it be easier, if it will be created automatically.

PR #1022, opened 2025-10-27
Removed setup.py, improved pyproject

as3ii:
- removed `pyinstaller` since it's not used anymore
- moved `sqlalchemy` to the right place (from `build` to required dependency)
- added `wheel` and `build` to the `build` dependencies
- switched from `requirements.txt` to the new python dependency handling via `pyproject.toml` (this reduces the installed dependencies to only those actually required)
  - `pip install .` installs only the libraries required to run freedata
  - `pip install .[test]` installs the libraries required for running and testing freedata
  - `pip install .[dev]` installs the libraries required for running, testing and linting/checking freedata
  - `pip install .[build]` installs only the libraries required to run and build the freedata pip module
  - `pip install .[nuitka]` installs only the libraries required to run and build the freedata executable
  - `requirements.txt` remains (temporarily?) for backward compatibility

This completes the migration from `setup.py` to `pyproject.toml` and should allow you to distribute freedata via pypi correctly, but further testing may be required

### DIFF
--- a/.github/workflows/build_appimage_bundle.yml
+++ b/.github/workflows/build_appimage_bundle.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install .[nuitka]
 
       - name: Build binaries
         working-directory: freedata_server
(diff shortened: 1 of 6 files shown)


DJ2LS (approved):

as3ii:
> We have a merge conflict, if this is solved we can merge :)

rebased over current develop branch 👍🏻

Issue #1108, opened 2026-07-23
[Bug]: /modem/send_arq_raw has never worked: always 503, then a crash, then success is invisible


spinkham:
I've been wiring up an automated interop test rig that drives two
FreeDATA instances over a simulated audio channel, and I wanted the raw
ARQ API as the transport. That turned into a small debugging story, so
I'm filing it as one issue with the whole picture. I have two small PRs
ready, linked below.

**1. The endpoint always returns 503 "Modem Busy"**

`post_arq_raw` checks `if ctx.state_manager.is_modem_busy:`, but
`is_modem_busy` is a `threading.Event`, so the object itself is always
truthy and every request gets a 503 regardless of what the modem is
doing. The check the rest of the codebase uses is `getARQ()` (the
beacon, `command_message_send.py`, `frame_handler_cq.py`). One gotcha
for whoever touches this next: `is_modem_busy.is_set()` would be wrong
in the *other* direction, because `setARQ(True)` clears the event
rather than setting it. `getARQ() == True` means busy. (The `getARQ`
docstring says the opposite of what the code does, which didn't help.)

**2. With that fixed, it crashes**

The endpoint enqueues `command_arq_raw.SendARQRawCommand`, which
doesn't exist anywhere in the tree; the class is `ARQRawCommand`. The
503 bug was masking this AttributeError completely, which I think means
this endpoint has never actually run for anyone.

**3. With both fixed, the transfer works, but you can't tell**

The over-the-air transfer now succeeds. Both stations log "All data
received successfully!", 512/512 bytes, checksum verified, state ENDED.
But the receiving side never emits `arq-session-finished` on success.
The abort path sends it, the failure path sends it, the sender's
success path sends it. Only the IRS success branch returns without one.
So no websocket or REST consumer can detect a successful inbound raw
transfer, and since `handle_raw` just returns the payload up a call
chain that discards it, the received data isn't reachable either.

I dug through the history because this seemed too basic to be missing
on purpose, and it looks like a regression: c2388a65 ("bringing back
statistics", April 2 2024) has the success branch emitting
`send_arq_session_finished(False, id, dxcall, True, state.name,
data=received_data, statistics=...)`, and by ef18f4cc (April 7 2024)
it's gone, even though that commit's diff of `arq_session_irs.py`
doesn't touch those lines. My best guess is it got lost in a merge.
Nobody noticed for two years because p2p messages notify through
`message_received` instead, and raw ARQ couldn't be reached from the
[...]

PR #1112, opened 2026-07-23
Lower RX audio latency and make the input callback real-time safe

spinkham:
This grew out of the same test rig as #1108: two FreeDATA instances
talking over a simulated audio channel (ALSA snd-aloop). On that rig
FreeDATA was completely deaf: `[AUDIO STATUS] input overflow` every
~408 ms from the moment the input stream opened, before any traffic,
and not one frame ever decoded. Digging into how the RX stream
negotiates its buffers turned into this PR. Three commits, RX side
only.

**What's in it**

1. **Run the RX DSP off the real-time input callback.** The input
   callback used to do the 48→8 kHz resample, the spectrum FFT, and the
   demod buffer pushes inline. It now copies the block onto a queue and
   returns; a worker thread does the DSP. If the queue is ever full the
   callback drops the block and counts it instead of blocking, so
   Python-side work can never stall the audio thread. The threading
   model is otherwise unchanged (RX DSP already ran off the TX thread).

2. **A unit test for that split**: the worker really performs the
   relocated DSP, and a full queue means a counted drop with the
   callback returning immediately. Needs no audio hardware.

3. **`blocksize=0` plus an explicit `latency=0.2` on the input
   stream.** This is the interesting one; numbers below.

**Why blocksize=0 + latency=0.2**

With `blocksize=4800`, PortAudio hands the demodulator fixed 100 ms
blocks, so received audio waits up to 100 ms in the input buffer before
decode can start, and that price is paid again at every turnaround of a
stop-and-wait ARQ session. `blocksize=0` lets PortAudio deliver small
blocks and cuts that to the 10–50 ms range.

But reading back the actually-negotiated ALSA parameters (from
`/proc/asound/.../hw_params` while the stream is open) showed the
defaults land badly in two different ways, depending on device:

| stream config | negotiated ring on snd-aloop | behavior |
|---|---|---|
| `blocksize=4800`, default latency | 9600 frames = 2 × 100 ms periods | overflow every 408 ms from stream open; modem deaf |
| `blocksize=0`, default latency | 1920 frames = 5 × 8 ms | clean, but only 40 ms deep |
| `blocksize=0`, `latency=0.2` | 12000 frames = 5 × 50 ms | clean, 250 ms deep |

(Identical numbers on two different machines.)

The stock configuration's failure on the loopback is structural, not
load: sounddevice's default "high" latency maps to 0.1 s there, which
at 100 ms periods is a plain double buffer, and a timer-driven virtual
device misses that service deadline on a fixed beat. It overflows with
the DSP inline or on the worker, idle or busy, and the capture side
ends up delivering only about half of real time; the modem never
decodes anything. Anyone running FreeDATA against a virtual audio cable
(automated testing, channel simulators) hits this as "RX never works".

Real hardware is a different story, which is why this isn't just
`blocksize=0`: on a Digirig CM108B, `blocksize=4800` negotiated a
200 ms ring and never overflowed, while `blocksize=0` alone shrank the
ring to ~40 ms, less margin against a short processing stall than
stock had. The explicit `latency=0.2` keeps the depth (250 ms measured
on the loopback; the CM108's own limit is 2 s so the request has room)
while keeping the small blocks. Both axes improve; nothing is traded
away.

**Validation**

- ruff check and `ruff format --check --preview` clean; the new test
  passes.
- End to end over the loopback rig: stock develop transfers 0 bytes
  (deaf, as above); this branch completes 4 KB raw-ARQ transfers intact
  at ~100 B/s on a clean simulated channel.
- The callback split alone was A/B'd against develop earlier with no
  behavior difference on either machine class; it's the enabler for
  small blocks more than a standalone fix.

If you'd rather have the latency exposed in config.ini instead of a
constant, or prefer a different depth, happy to rework. 0.2 s is the
smallest round value that keeps the measured ring at or above what
`blocksize=4800` negotiated on real hardware.

### DIFF
--- a/freedata_server/modem.py
+++ b/freedata_server/modem.py
@@ -6,6 +6,7 @@ Created on Wed Dec 23 07:04:24 2020
 """
 
 import queue
+import threading
 import time
 from freedata_server import codec2
 import numpy as np
@@ -75,6 +76,15 @@ class RF:
 
         self.data_queue_received = queue.Queue()
 
+        # RX audio captured by the real-time sounddevice callback is handed to
+        # rx_audio_processing_worker through this queue, keeping the callback
+        # minimal (copy + enqueue). Running the DSP in the callback under the GIL
+        # is what makes it miss its deadline and overflow on slower CPUs.
+        self.rx_audio_in_queue = queue.Queue(maxsize=100)
+        self.rx_audio_worker_running = False
+        self.rx_audio_worker_thread = None
+        self.rx_audio_dropped_blocks = 0
+
         self.demodulator = demodulator.Demodulator(self.ctx)
         self.modulator = modulator.Modulator(self.ctx)
 
@@ -116,6 +126,12 @@ class RF:
             # self.stream = lambda: None
             # self.stream.active = False
             # self.stream.stop
+            # stop the RX audio processing worker before closing the streams
+            self.rx_audio_worker_running = False
+            try:
+                self.rx_audio_in_queue.put_nowait(None)
+            except queue.Full:
+                pass
             self.sd_input_stream.close()
             self.sd_output_stream.close()
         except Exception as e:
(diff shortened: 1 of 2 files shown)


DJ2LS:
Thanks for the PR! 

I'm carefully regarding audio. blocksize=4800 has been a reliable working tradeoff. So if we want to change it, I suggest keeping 4800 the default for now with an option in settings to change it down to fixed and validated values which allow lower latency.

PR #1115, opened 2026-07-25
fix(audio): re-block captured RX audio into whole DSP blocks (fixes the AssertionError from #1112)

spinkham:
Follow-up to #1112, fixing the `[AUDIO EXCEPTION] e=AssertionError()` you hit after merging it. Sorry for the breakage, and thanks for the quick report. One commit.

**Root cause**

`codec2.resampler.resample48_to_8` asserts its input length is a multiple of 6 (`FDMDV_OS_48`). With `blocksize=0`, PortAudio picks the capture block size per device. The devices I measured for #1112 (CM108, snd-aloop) happened to deliver multiple-of-6 blocks, but your device likely delivers something else like a 512/1024-class size:

```
 4800 frames -> OK               4800 % 6 = 0
  512 frames -> AssertionError    512 % 6 = 2
 1024 frames -> AssertionError   1024 % 6 = 4
```

So every captured block raised, the DSP chain never ran, and RX was completely deaf. The failure was total, not intermittent, and the old test never caught it because it captured 4800-frame blocks, which are a multiple of 6.

**Fix**

The capture block size and the DSP block size are now decoupled. The worker appends whatever the callback captured to a carry buffer and runs the DSP chain once per whole 4800-sample block available; the remainder carries into the next captured block. The sample stream stays gapless (the resampler's filter memory spans blocks) and the DSP always receives exactly 4800 samples, no matter what the device delivers.

On your comment in #1112 about keeping 4800 as the reliable tradeoff: that is effectively what this does, on every device. The DSP side is pinned to a named constant `RX_DSP_BLOCK_48K = 4800`, with a comment listing the four places that depend on it. Only the capture side stays at `blocksize=0` + `latency=0.2`, because a fixed `blocksize=4800` is the configuration that negotiates a two-period ring on snd-aloop and drops audio continuously (the deaf-on-loopback case from #1112). With the re-blocking in place the captured size no longer matters for correctness, so if you would still like a settings option for the capture blocksize, it is now safe to add on top of this: any value works.

Re-blocking also quietly fixes three degradations that short blocks were causing (short blocks fed the FFT/channel-busy path mostly zero padding, the audio streaming queue got one zero-padded chunk per block regardless of size, and the per-block auto level got faster and jumpier the smaller the block).

**Tests**

- New `TestRxAudioReblocking` class, and the capture size in the existing tests changed from 4800 to 512 so the suite exercises the failing shape. Covers: odd capture sizes never reach the resampler; exact block accounting including the carried remainder; the sample stream is preserved bit-exactly (nothing dropped, duplicated or reordered); and every re-blocked block is accepted end to end by the real codec2 resampler, which is the actual regression guard for this bug.
- Against the pre-fix code, 5 of the 6 tests fail; with the fix, all pass.
- `ruff check` and `ruff format --check` clean.

Not yet verified on your hardware since the exact block size there is unknown, but the assertion is now unreachable from the capture path regardless of size. If you still see `AUDIO EXCEPTION` after this, please grab the log line and I will dig further.

### DIFF
--- a/freedata_server/modem.py
+++ b/freedata_server/modem.py
@@ -71,6 +71,26 @@ class RF:
         self.AUDIO_STREAMING_CHUNK_SIZE = 2400
         self.audio_out_queue = queue.Queue()
 
+        # Size of the block the RX DSP chain runs on, in 48 kHz samples. The input
+        # stream is opened with blocksize=0 (PortAudio picks the capture size per
+        # device), so the DSP must not rely on the captured block size:
+        # rx_audio_processing_worker re-blocks whatever the callback is handed
+        # to exactly this size, so a device or setting that delivers some other
+        # size cannot reach the DSP. 4800 (100 ms, 800 samples at 8 kHz) is the
+        # size the rest of the chain is built around:
+        #   * codec2.resampler.resample48_to_8 asserts len % FDMDV_OS_48 (6) == 0
+        #     and raises AssertionError otherwise,
+        #   * audio.calculate_fft pads/truncates to 800 samples at 8 kHz, so a
+        #     shorter block is mostly zero padding and the spectrum, channel busy
+        #     detection and audio_dbfs all degrade,
+        #   * enqueue_streaming_audio_chunks zero-pads every block up to
+        #     AUDIO_STREAMING_CHUNK_SIZE, so a shorter block streams mostly
+        #     silence and emits one chunk per block regardless of size,
+        #   * audio.normalize_audio (rx_auto_audio_level, on by default)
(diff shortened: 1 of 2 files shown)


spinkham:
Follow-up validation: ran this end to end over the snd-aloop loopback rig (two server instances, half-duplex simulated channel, same setup as the #1112 numbers). Two clean-channel 4096-byte raw-ARQ transfers, both completed byte-for-byte intact at 97.7 and 105.3 B/s, right on the ~100 B/s baseline from before. So the re-blocking does not disturb the working loopback path; the unit tests already cover the odd-block-size hardware case that broke for you.

